PR #1, opened 2021-08-09
style: format using Google style

hei-ryan:


.github/workflows/quality.yml
```diff
@@ -12,43 +12,43 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
```
hei-teacher, line 15: CI checks didn't work because you modified the CI file `quality.yml`. In the future, do **NOT** modify files under `.github/workflows/` without prior discussion.

Since your proposed formatting is indeed the good one, I applied the change directly into `dev`. Please rebase. The CI checks should execute after your rebase.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 <a name="0.1.0"></a>
-## 0.1.0 (2021/08/08)
```
hei-teacher, line 2: This shouldn't be changed. Check if your line endings is in Linux style (LF) instead of Windows style (CRLF)
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

LICENSE.md
```diff
@@ -2,21 +2,16 @@ The MIT License (MIT)
 
 Copyright (c) 2021 Haute École d'Informatique (HEI) de Madagascar
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
```
hei-teacher, line 5: Revert this file formatting to its previous state. Indeed your commit is only about removing the warnings related to Google check. The LICENSE file has nothing to do with that task. Remember to only do in your commit what your commit title suggests, and nothing more.

build.gradle
```diff
@@ -1,72 +1,72 @@
 plugins {
-	id 'java'
-	id 'org.springframework.boot' version '2.5.3'
-	id 'io.spring.dependency-management' version '1.0.11.RELEASE'
+    id 'java'
```
hei-teacher, line 2: I like this formatting. Normally it does not affect the Google style warnings, but it's in the spirit, let's keep it.

config/checkstyle/google_checks.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
-          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
```
hei-teacher, line 3: Revert this file to its previous state.

hei-teacher (changes requested):

hei-teacher (approved):

PR #4, opened 2021-08-11
docs/ha-api-spec

hei-ryan:
According to the last comments, the endpoints have been simplified. However, the three endpoints weren't enough to describe their actions.
Some modifications have been made in the schemas so they can be compatible with the paths.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -1,0 +1,1079 @@
+openapi: "3.0.3"
+info:
+  version: "1.0"
+  title: "HEI Admin API"
+paths:
+  /groups:
+    get:
+      tags:
+        - "Groups"
+      summary: "Find groups based on search criteria"
+      parameters:
+        - name: "id"
+          in: "query"
+          description: "Identifier of the group to find"
+          required: false
+          schema:
+            type: string
+        - name: offset
+          in: query
+          required: false
+          description: "Zero-based offset for pagination. Defaults to `0`."
+          schema:
+            type: integer
+        - name: limit
+          in: query
+          required: false
+          description: "Pagination limit, that is the maximum number of elements that can be retrieved. Defaults to `1000` and cannot exceed `1000`."
+          schema:
+            type: integer
+
+      operationId: "findGroups"
+      responses:
+        200:
+          description: "List of groups"
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Group'
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+    post:
+      tags:
+        - "Groups"
+      summary: "Create new groups"
+      operationId: "createGroups"
+      requestBody:
+        description: "Collection of groups to create"
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateGroup'
+      responses:
+        200:
+          description: "Collection of created groups"
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Group'
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+    patch:
+      tags:
+        - "Groups"
+      summary: "Update existing groups"
+      description: |
+
+        Possible udpates on:
+        * Name: change
+
+      operationId: "updateGroups"
+      requestBody:
+        description: "Groups to update"
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Group'
+            examples:
+              changeName:
+                summary: Change name
+                value:
+                  resources:
+                    - id : "uuid"
+                    - id : "uuid"
+                  name: "Groupe A"
+      responses:
+        200:
+          description: "The updated groups"
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Group'
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+    delete:
+      tags:
+        - "Groups"
+      summary: "Delete existing groups"
+      operationId: "deleteGroups"
+      parameters:
+        - name: "ids"
+          in: "query"
+          description: "Identifiers of the groups that need to be deleted separated by coma"
+          example:
+            "uuid,uuid"
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          $ref: "#/components/responses/200Delete"
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+  /students:
+    get:
+      tags:
+        - "Students"
+      summary: "Find students based on search criteria"
+      description: "  Please notice that CourseHours are generated automatically, based on the availabilites of the teachers."
+      parameters:
+        - name: "id"
+          in: "query"
+          description: "Identifier of the student to find"
+          required: false
+          schema:
+            type: string
+        - name: offset
+          in: query
+          required: false
+          description: "Zero-based offset for pagination. Defaults to `0`."
+          schema:
+            type: integer
+        - name: limit
+          in: query
+          required: false
+          description: "Pagination limit, that is the maximum number of elements that can be retrieved. Defaults to `1000` and cannot exceed `1000`."
+          schema:
+            type: integer
+
+      operationId: "findStudents"
+      responses:
+        200:
+          description: "List of students"
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Student'
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+    post:
+      tags:
+        - "Students"
+      summary: "Create new students"
+      description: ""
+      operationId: "createStudents"
+      requestBody:
+        description: "Collection of students to create"
+        required: true
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/CreateStudent'
+      responses:
+        200:
+          description: "Collection of created students"
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Student'
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+    patch:
+      tags:
+        - "Students"
+      summary: "Update existing students"
+      description: |
+        Possible udpates on:
+        * Group: change
+        * Status: change
+        * EnrolledCourses: add
+        * Grades : add
+      operationId: "updateStudents"
+      requestBody:
+        description: "Students to update"
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Student'
+            examples:
+              addGroup:
+                summary: Change group
+                value:
+                  resources:
+                    - id : "uuid"
+                    - id : "uuid"
+                  group:
+                    id: "uuid"
+              changeStatus:
+                summary: Change status
+                value:
+                  resources:
+                    - id : "uuid"
+                    - id : "uuid"
+                  status: "ENABLED"
+              addEnrolledCourse:
+                summary: Add enrolled course
+                value:
+                  resources:
+                    - id : "uuid"
+                    - id : "uuid"
+                  course_id: "uuid"
+                  course_name: "Architecture logicielle"
+              addGrade:
+                summary: Add grade
+                value:
+                  resources:
+                    - id : "uuid"
+                    - id : "uuid"
+                  value: "0"
+                  enrolled_course_id: "uuid"
+                  enrolled_course_name: "Architecture logicielle"
+
+      responses:
+        200:
+          description: "The updated Students"
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Student'
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+    delete:
+      tags:
+        - "Students"
+      summary: "Delete existing students"
+      operationId: "deleteStudents"
+      parameters:
+        - name: "ids"
+          in: "query"
+          description : "Identifiers of the students that need to be deleted separated by coma"
+          required: true
+          example:
+            "uuid,uuid"
+          schema:
+            type: string
+      responses:
+        200:
+          $ref: "#/components/responses/200Delete"
+        400:
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+  /courses:
+    get:
+      tags:
+        - "Courses"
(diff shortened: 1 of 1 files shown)


docs/api.yaml
```diff
@@ -0,0 +1,641 @@
+openapi: "3.0.3"
+info:
+  description: |
+    * To complete
```
hei-teacher, line 4: Just no description for the moment

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+
+  version: "1.0.0"
+  title: "HEI Admin API"
+  contact:
+    email: "contact@hei.school"
+
+servers:
+  - description: To complete
```
hei-teacher, line 12: Just no server for the moment

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+  /impersionation:
```
hei-teacher, line 274: No. Impersonation is about calling some endpoints of the API while faking being someone else. Typically we want staff people to be able to call the GET endpoints (and only the GET endpoints as we want impersonation to be read-only) while pretending for example to be logged as a specific teacher or a specific student. I don't feel like we need to create a specific endpoint for doing this. Just proper definition of the STAFF role at implementation level.

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+  /activation:
```
hei-teacher, line 242: No. PATCH the Student resource, and the PATCH-ed attribute is status. That is: `PATCH /students/{id}` with payload `{'status' : 'ENABLED'}`

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+  /group:
```
hei-teacher, line 146: No. PATCH the Student resource, and the PATCH-ed attribute is group. That is: PATCH /students/{id} with payload {'group_id' : 'uuid'}

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+    ######################################################
+
+    Grade:
+      type: object
+      properties:
+        id:
+          type: string
+        value:
+          type: integer
+        enroledCourseId:
```
hei-teacher, line 356: Put every attribut in snake_case instead of camelCase

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+  /courses/student:
```
hei-teacher, line 178: No. PATCH the student resource

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+  /courses/teacher:
```
hei-teacher, line 210: No. Patch the teacher resource

docs/api.yaml
hei-teacher, line 356: enroled --> enrolled

docs/api.yaml
```diff
@@ -0,0 +1,641 @@
...
+          $ref: "#/components/responses/400"
+        403:
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
```
hei-teacher, line 305: Again. These endpoints are too complicated. You totally forgot the `/students` and `/teachers` endpoints even though they are the central resources of the API. Please specify the following calls in priority and see what we can do with them before thinking about any additional endpoint:

- GET /students (for search with filter parameters, eg, by id, by date, by anything)
- POST /students (for batch creation)
- PATCH /students (for batch update)
- GET /teachers
- POST /teachers
- PATCH /teachers

hei-teacher (changes requested):

docs/api.yaml
hei-ryan: The parameter id is not inside the path but the query.

docs/api.yaml
hei-ryan: Is the group_id enough for this request ? 
Or is it the same as if I had all the groupe attribute (id and name) ?

docs/api.yaml
hei-ryan: Adding an enrolled course to a student PATCH-ed using 'course_id' and other necessary attributes.

docs/api.yaml
hei-ryan: Same as the student's.

docs/api.yaml
hei-ryan: The endpoints are now changed to 4 endpoints : students,courses,teachers and grades.
I noticed that resources of an object with a relation ONE TO ONE/MANY can be easily added with PATCH verb but when I try with an object with a relation MANY TO MANY,  it's not the same so that's why I created the endpoint "grades", using a POST verb.

docs/api.yaml
hei-teacher: Thinking again about it, do not put the id neither in the path nor in the parameters. Doing so will prevent batch patching (ie, patching a whole list of students). You should put the id the in payload, attached to each student we want to patch.

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+  title: "HEI Admin API"
+  contact:
+    email: "contact@hei.school"
+
+  #servers:
+  # - description: To complete
+  #   url: example.com
+
+paths:
+  /grades:
```
hei-teacher, line 15: Not needed. A grade should be an attribute of a student.enrolled_course. When not yet graded, student.enrolled_course.grade should be null. When giving grade, just patch that attribute

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+    post:
+      tags:
+        - "Teacher"
+      summary: "Create new Teachers"
```
hei-teacher, line 345: "Teachers" --> "teachers" (not capitalized)

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+  /courses:
+    get:
+      tags:
+        - "Course"
+      summary: "Find courses with optionnal search criteria"
```
hei-teacher, line 205: Find courses with... --> Find courses based on search criteria. Also update other corresponding definition for students and teachers

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+    post:
+      tags:
+        - "Course"
+      summary: "Create new Courses"
```
hei-teacher, line 249: "C" --> "c"

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+          $ref: "#/components/responses/403"
+        404:
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+    post:
+      tags:
+        - "Course"
```
hei-teacher, line 248: Put tags to plural: Students, Teachers, Courses, Groups

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+        - "Teacher"
+      summary: "Find teachers with optionnal search criteria"
+      parameters:
+        - name: "id"
+          in: "query"
+          description: "Identifier of the teacher to find"
+          required: true
+          schema:
+            type: string
+        - name: from
```
hei-teacher, line 293: From and to filters are not needed for teachers. There aren't that many, let consumers handle that

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+        - "Student"
+      summary: "Find students with optionnal search criteria"
+      parameters:
+        - name: "id"
+          in: "query"
+          description: "Identifier of the student to find"
+          required: false
+          schema:
+            type: string
+        - name: from
```
hei-teacher, line 59: These from/to parameters are not relevant in this API. What we might want is some parameter for filtering the timetable attribute, as it will grow through time and will eventually become huge. We can imagine some attribute like `futureTimetableOnly: boolean` or  `timetableForCourses: list_of_courses_uuid`. But let's wait for the need to arise. For the moment, just remove from/to.

docs/api.yaml
```diff
@@ -0,0 +1,844 @@
...
+      parameters:
+        - name: "id"
+          in: "query"
+          description: "Identifier of the student to update"
+          required: true
+          schema:
+            type: string
+      description: |
+        Possible udpates on:
+        * Groups: change
```
hei-teacher, line 155: Groups --> group (singular)

hei-teacher (changes requested):

docs/api.yaml
hei-ryan: Adding a grade is now a PATCH-ed action in student endpoint.

docs/api.yaml
hei-ryan: Only id used, fixed for all similars actions except for those who have specified other attributes, eg : main_teacher_name in course.

DELETE verb added to other endpoints. Please comment the 200 Response if needed before resolving this conversation.

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
...
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+    patch:
+      tags:
+        - "Groups"
+      summary: "Update an existing group"
+      parameters:
```
hei-teacher, line 96: Add the creation_date parameter

hei-teacher, line 96: Add a (ref: string) parameter, described as follows: "Human-readable identifier for the resource". Typically, ref is auto-incrementable field which will be shown afterwards on all user interfaces.

On the other hand, id is an uuid, which is NOT auto-increment-able. As search are based on id, we MUST make it non-autoincrementable (so that malicious people cannot guess our range of id).

Please also add ref to all other resources. Note that ref is NOT updatable.

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
...
+          enum:
+            - ENABLED
+            - DISABLED
+        group:
+          $ref: '#/components/schemas/Group'
+        enrolled_courses:
+          type: array
+          items:
+            $ref: '#/components/schemas/EnrolledCourse'
+        grades:
```
hei-teacher, line 758: NO. grade is just an attribute of  an enrolled_course. Just add grade to the objects of the enrolled_courses list...

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
...
+          type: integer
+        timetable:
+          type: array
+          items:
+            $ref: "#/components/schemas/TaughtPeriod"
+
+    EnrolledCourse:
+      type: object
+      properties:
+        id:
```
hei-teacher, line 973: Do not put id here. Just course_id is enough. In a REST API, when you give an id, you should also give an endpoint for manipulating the identified resources. We don't need any /enrolledCourses endpoint. Similarly, also remove taughCourse.id, just use course_id

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
...
+      parameters:
+        - name: "id"
+          in: "query"
+          description : "Identifier of the student that needs to be deleted"
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: "Student deleted"
```
hei-teacher, line 323: No. Be more generic to allow re-usability. Create a new 200Delete response described as "Resources deleted" (mind the plural)

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
...
+          type: "string"
+          format: "date"
+        hire_date:
+          type: "string"
+          format: "date"
+        taught_courses:
+          type: array
+          items:
+            $ref: "#/components/schemas/TaughtCourse"
+        interventions:
```
hei-teacher, line 843: What's the use of this? What's the difference with TaughtCourse.timetable?
Again, student.timetable is the class hours student must attend to. teacher.timetable is about the class hours the teacher must deliver. course.timetable is the union of all teacher.timetable

Again, if you're not comfortable with using the same name but having different meaning depending on the context, prefix the name with some context

hei-teacher (changes requested):

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
+openapi: "3.0.3"
+info:
+  #  description:
```
hei-teacher, line 3: remove

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
...
+
+  version: "1.0.0"
+  title: "HEI Admin API"
+  contact:
+    email: "contact@hei.school"
+
+  #servers:
```
hei-teacher, line 10: remove

hei-teacher (changes requested):

docs/api.yaml
hei-ryan: Do you mean the comments ?

docs/api.yaml
hei-ryan: Can a course own many grades ? 
For example : 
When a student fails, is it necessary to have a history of the grades by date, so we create an object of type date with its value and its date and choose the most recent.

Or

We just update the grade value so the grade isn't an object but just an attribute of type integer ?

docs/api.yaml
hei-ryan: Ok.
This conversation can be resolved, the rest is in the TaughtCourse usage's conversation.

docs/api.yaml
hei-ryan: The attribute required_teaching_hours is the difference with the TaughtCourse and the TaughtCourse.timetable (TaughtPeriod).
As you said, a course has a main teacher, but other teachers can give the same course so when you added this attribute I think it's to know how many hours is required by each teachers.

I do agree with you with the course.timetable, but we didn't add this attribute _(yet)_ because this feature has not _(yet)_ been defined in the administration profile.

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
...
+          $ref: "#/components/responses/404"
+        429:
+          $ref: "#/components/responses/429"
+        500:
+          $ref: "#/components/responses/500"
+
+    patch:
+      tags:
+        - "Groups"
+      summary: "Update an existing group"
```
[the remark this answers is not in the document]
hei-ryan: So we use a collection of ids not a singular id in required parameters, and do the actions (UPDATE or DELETE) of all of the objects corresponding to these ids in a SINGLE request ?

I just want to be sure this is what you are waiting for.

docs/api.yaml
hei-ryan: I actually removed creation_date attribute according to the past conversation about the from/to parameter not needed yet, for ALL schemas.

Can you explain what is it use for ?

hei-ryan: Is there a specific or general standard to write this ref according to the alphabetic and numeric characters ?
Or this detail isn't important in Rest API ?

If it is : 
How many alphabetic chars ? How do we "split" it ? Do we use underscore or not ? How many digits for the numeric part ?
For example for student : "STD_00001" or "STD000001".

hei-teacher: creation_date is useful for identifying to which "promo" a group belongs to. We just don't need a search filter on that attribute since there are not that many groups. It's good that the search filter was removed, but keep the information nethertheless.

hei-teacher: No standard way, it's actually business-dependent

docs/api.yaml
hei-teacher: excellent point. we indeed want to keep trace of the number of attempts before succeeding. Hence, instead ot the enrolled_course.grade, put enrolled_course.exams

docs/api.yaml
hei-teacher: yeah, never put "draft" comments in code

docs/api.yaml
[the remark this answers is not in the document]
hei-teacher: Not in parameters. But in paylaod. Only the GET verbs should have parameters. For all other verbs, put everything in payload. Eg for DELETE: `{'resources': [{'id': '1'}, {'id': '2'}]}`

docs/api.yaml
```diff
@@ -0,0 +1,1061 @@
+openapi: "3.0.3"
```
hei-teacher, line 1: Add a `README.md` to the repository in which you indicate that the API can be browsed at https://petstore.swagger.io/?url=https://raw.githubusercontent.com/hei-school/hei-admin-api/docs/ha-api-spec/docs/api.yaml

That link renders the yaml definition into a read-only html (user-friendly) file

hei-teacher (changes requested):

hei-teacher (approved):

PR #5, opened 2021-08-17
feat/auth

hei-ryan:
Authenticate the REST API calls using JWT provided by Amazon Cognito.
When the calls are authenticated, the server-side (Spring-boot) should identify the profile of the client.


hei-ryan:
For the profile, could I just add a custom attribute in User Pool, eg : profile, and just use this attribute in Spring after I get the token ?

application.yml
```diff
@@ -0,0 +1,6 @@
+com:
+  kb:
+    jwt:
+      aws:
+        userPoolId: "us-east-2_ijXLzbX6J"
```
hei-teacher, line 5: Configure, or remove

application.yml
```diff
@@ -0,0 +1,6 @@
...
+        region: "us-east-2"
```
hei-teacher, line 6: Configure, or remove

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
```diff
@@ -0,0 +1,37 @@
+package school.hei.haapi.authentication;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Component;
+
+import javax.servlet.*;
```
hei-teacher, line 10: Explicit all your imports

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+@Component
+public class AwsCognitoJwtAuthFilter extends GenericFilter {
+
+    private static final Log logger = LogFactory.getLog(AwsCognitoJwtAuthFilter.class);
```
hei-teacher, line 16: 1. Caps for constans
2. Use @slf4j of Lombok

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
```diff
@@ -0,0 +1,37 @@
...
+    @Autowired
+    private AwsCognitoJwtIdToken cognitoIdTokenProcessor;
```
hei-teacher, line 18: 1. No setter/attribute injection. Use constructor injection https://dzone.com/articles/spring-di-patterns-the-good-the-bad-and-the-ugly
2. Use final attributes as much as possible (and more genarally, immutable value)

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public AwsCognitoJwtAuthFilter(AwsCognitoJwtIdToken cognitoIdTokenProcessor) {
```
hei-teacher, line 20: Use @AllArgsConstructor

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
```diff
@@ -0,0 +1,37 @@
...
+        this.cognitoIdTokenProcessor = cognitoIdTokenProcessor;
+    }
```
hei-teacher, line 22: Empty line missing

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain filterChain) throws IOException, ServletException {
+        Authentication authentication;
+        try {
+            authentication = this.cognitoIdTokenProcessor.authenticate ((HttpServletRequest)request);
+            if (authentication != null) {    SecurityContextHolder.getContext().setAuthentication(authentication);
+            }
+        } catch (Exception var6) {
```
hei-teacher, line 30: Use meaningful names

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.security.core.userdetails.User;
+import org.springframework.stereotype.Component;
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.Collection;
+import java.util.List;
```
hei-teacher, line 16: Empty line

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+  private String getUserNameFrom(JWTClaimsSet claims) {
+    return claims.getClaims()
+            .get(this.jwtConfiguration.getUserNameField())
+            .toString();
+  }
+
+  private void verifyIfIdToken(JWTClaimsSet claims) throws Exception {
+    if (!claims.getIssuer().equals(this.jwtConfiguration.getCognitoIdentityPoolUrl())) {
+      throw new Exception("Not an ID Token");
```
hei-teacher, line 52: 1. Never use bare Exception or bare RuntimeException. Always extends them and use more specific exceptions.
2. Add an interceptor for handling in a uniform way the mapping from interal exception to external Error Model. See ExceptionMapper (return ApiException in the different mapping methods)

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+
+  private void validateIssuer(JWTClaimsSet claims) throws Exception {
+    if (!claims.getIssuer().equals(this.jwtConfiguration.getCognitoIdentityPoolUrl())) {
+      throw new Exception(String.format("Issuer %s does not match cognito idp %s", claims.getIssuer(), this.jwtConfiguration.getCognitoIdentityPoolUrl()));
```
hei-teacher, line 58: Limit the use of checked exceptions. Use unchecked exceptions as much as possible

src/main/java/school/hei/haapi/authentication/JwtAuthentication.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.nimbusds.jwt.JWTClaimsSet;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.core.GrantedAuthority;
+
+import java.util.Collection;
+public class JwtAuthentication extends AbstractAuthenticationToken {
+
+    private final Object principal;
+    private JWTClaimsSet jwtClaimsSet;
```
hei-teacher, line 11: final

src/main/java/school/hei/haapi/authentication/JwtProcessor.java
```diff
@@ -0,0 +1,33 @@
...
...PICPEVWh94AeyBLycjYdJG-Wue8FuMYSxjIht6dPE5TDtA6bTQulc9kkQqzCdLbhQlEJzDG-fQxQntrIqx3ohn16NIG_IJxxBmRumB71suiJukAifHNdGqsSjAeOj7HEa9V-Q80zG8W2ue1I3FnDpO0qlCXfHHTalmPss3UCYsdkandV6zeUGalqEWiGBpM4EgwH7tkQ3Gqy1Cw&access_token=eyJraWQiOiJSN2szTEVkZ3ZJUVhWaVVvMEhuTWJUVTAwSGplM2l3b2duS0hPR1ZmdzJ3PSIsImFsZyI6IlJTMjU2In0.eyJzdWIiOiI5ZmViNTMzZC01ODA0LTQ0MzgtOTU0YS01OTk3YjFiNGM1OWQiLCJ0b2tlbl91c2UiOiJhY2Nlc3MiLCJzY29wZSI6ImF3cy5jb2duaXRvLnNpZ25pbi51c2VyLmFkbWluIHBob25lIG9wZW5pZCBwcm9maWxlIGVtYWlsIiwiYXV0aF90aW1lIjoxNjI5MjAyNjIzLCJpc3MiOiJodHRwczpcL1wvY29nbml0by1pZHAudXMtZWFzdC0yLmFtYXpvbmF3cy5jb21cL3VzLWVhc3QtMl9palhMemJYNkoiLCJleHAiOjE2MjkyMDYyMjMsImlhdCI6MTYyOTIwMjYyMywidmVyc2lvbiI6MiwianRpIjoiMGM2ODMzMjUtZTM0ZC00YzkyLTgxNjYtMzg5NjgxYzY5YmY2IiwiY2xpZW50X2lkIjoiN2ViOXZkNzM1N3QzN3I1MGJnNmhma3YwOTQiLCJ1c2VybmFtZSI6InJ5YW4ifQ.w4QWOVVWHXjjhTObq91Ufd1XgNJmttbifrG1X3HsIiTLH9UkQwJVM2ry9Orb8ck1ndPo5MAYhEAGKMlKiNyWC4IFZb-cqM7Jt-dkJh4_FlXOvvIwbvz14TZYadFTkkm1AgsuKxoB106gpVITEVwDqFcBYibqv7Ob2DuMFWs3CWk2Xlx1N8_wfkbOuhVF78zsmyie7I-vpQGDqll7LTqthzzMs5jfY0ljI6lQ8Z-2Pt59rhqhvfYlsgsJY0NfX_N_MoJG9WEpRTlbXMTaHqfrmObXMdBCIGAhjBtsxeOkiPS5mbl2G-y8SS0pso7KoJZM7g9zaLhX7Z00mGyJr4WTuQ&expires_in=3600&token_type=Bearer
```
hei-teacher, line 32: Remove that comment

src/main/java/school/hei/haapi/authentication/JwtProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+@Configuration
+public class JwtProcessor {
+    private static final JWSAlgorithm RS256 = new JWSAlgorithm("RS256");
+    @Autowired
+    private JwtConfiguration jwtConfiguration;
+    @Bean
+    public ConfigurableJWTProcessor configurableJWTProcessor() throws MalformedURLException {
+        ResourceRetriever resourceRetriever = new DefaultResourceRetriever(2000, 2000);
```
hei-teacher, line 25: No magic numbers

src/main/java/school/hei/haapi/authentication/SecurityConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+    @Override
+    protected void configure(HttpSecurity http) throws Exception {
+
+        http.headers().cacheControl();
+        http.csrf().disable()
+                .authorizeRequests()
+                .antMatchers("/ping").authenticated()
+                .anyRequest().authenticated()
+                .and()
+                .addFilterBefore(awsCognitoJwtAuthenticationFilter, UsernamePasswordAuthenticationFilter.class);
```
hei-teacher, line 24: UsernamePassword? NOT SURE

hei-teacher (approved):

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,67 @@
...
+
+@Component
+public class AwsCognitoJwtIdToken {
+  @Autowired private JwtConfiguration jwtConfiguration;
+  @Autowired private ConfigurableJWTProcessor configurableJWTProcessor;
+  private static final Log logger = LogFactory.getLog(AwsCognitoJwtIdToken.class);
+
+  private Collection<GrantedAuthority> getAuthorities() {
+    return List.of(
+        new SimpleGrantedAuthority("ROLE_USER"), new SimpleGrantedAuthority("ROLE_ADMIN"));
```
hei-teacher, line 26: On second thought, use Cognito for authentication only (so no role to add in the token it sends). The corresponding role will be stored by the API in its own database.

hei-teacher (changes requested):

application.yml
hei-ryan: The simple command **System.getenv("variable-name")** can read an environment variable.
Hence should I create/use an 'application.properties' or similar to inform the DevOps that this variable should be created/updated before the launch ?

src/main/java/school/hei/haapi/authentication/SecurityConfiguration.java
hei-ryan: After making some research, I found that UsernamePasswordAuthenticationFilter processes an authentication form submission (username/password), and when I searched for an alternative everyone who is using token especially jwt always use UsernamePassword...

hei-ryan:
Please can you confirm if Sonar uses the Google Java style (https://google.github.io/styleguide/javaguide.html)
 instead of checkstyle (https://checkstyle.sourceforge.io/) ?

Actually, Sonar sees more 4 consecutive capital letters for a variable name as violation, eg : USER_POOL_ID.
But when i check in the Google Java style above, it's even recommended.

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-ryan: So the users informations are duplicated in the API database ? including the id ?
Because if I understand your "second thought", we will use the id included in the token then the "local User" by using id to identify the role ?

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
hei-ryan: Please refer to my comment about the use of checkstyle instead of Google Java Style in Sonar.

application.yml
hei-teacher: `application.properties` and `application.yaml` actually are the same thing, but in different format only. Since we already use properties file, so just remove this yaml file.

In order to tell the DevOps that this property should be overiddent at runtime, add the following comment in application.properties:

`# kb.jwt.aws.userPoolId=`

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
hei-teacher: constant = static final

I see that you put in caps lock final variables. Please revert them to caps unlock. Only variables that are both static and final are called constants.

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-teacher: Just use NotAuthorizedException instead of the two exceptions you created

src/main/java/school/hei/haapi/authentication/SecurityConfiguration.java
hei-teacher: Not conviced. There should be some kind of DefaultAuthenticationFilter before which you can put your CognitoFilter. UserNameFilter is really irrelevant here since we don't use any form

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-teacher: Yeah, duplicated. This is the notion of primary source: we use the API's database as main source of information. And Cognito is just a sort of replica. Note that identifiers are different between API's db and Cognito's db.

Also note that the role is not forwarded to Cognito, only the user's [first]name and email

application.yml
hei-ryan: Generally, environment var is written in capital with underscore. So when the DevOps see this `kb.jwt.awt.userPoolId=` he understands that the variable name will be `USER_POOL_ID` ?

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
hei-ryan: Ah ok.
I was confused about the use of immutable value ...

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-ryan: And how to identify the exception ? And in what case do we create/use a specific exception ?

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-ryan: So the primary key to identify a user is the unique ([first]name,email) ?

src/main/java/school/hei/haapi/authentication/SecurityConfiguration.java
hei-ryan: My bad, the BasicAuthenticationFilter is enough.

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
hei-ryan: Configuration's attributes are now constant (static final) and beans are injected by constructor

application.yml
hei-teacher: yes, there is a naming convention for setting properties through env_var

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-teacher: no, there should be an additional uuid. anyway, that should be the responsibility of cognito. look at its api.

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,77 @@
...
+  public AwsCognitoJwtIdToken(
+      final JwtConfiguration jwtConfiguration,
+      final ConfigurableJWTProcessor<SecurityContext> configurableJwtProcessor) {
+    this.jwtConfiguration = jwtConfiguration;
+    this.configurableJwtProcessor = configurableJwtProcessor;
+  }
+
+  public Authentication authenticate(HttpServletRequest request) {
+    String idToken = request.getHeader(this.jwtConfiguration.getHttpHeader());
+    log.info("Token ID : " + idToken);
```
hei-teacher, line 37: NEVER log token information! If anyone access to our log, they could do bad things

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,77 @@
...
+    if (idToken != null) {
+      try {
+        JWTClaimsSet claims =
+            this.configurableJwtProcessor.process(this.getBearerToken(idToken), null);
+        validateIssuer(claims);
+        verifyIfIdToken(claims);
+        String username = getUserNameFrom(claims);
+        Collection<GrantedAuthority> authorities =
+            List.of(new SimpleGrantedAuthority("ROLE_ADMIN"));
```
hei-teacher, line 46: Use Enum

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,77 @@
...
+        if (username != null) {
+          User user = new User(username, "", authorities);
+          return new JwtAuthentication(user, authorities);
+        }
+      } catch (NotAuthorizedException | ParseException | BadJOSEException | JOSEException e) {
+        log.error(e.getMessage());
```
hei-teacher, line 52: Don't just log exceptions. Do additional things or don't just catch them.

src/main/java/school/hei/haapi/authentication/JwtAuthentication.java
```diff
@@ -0,0 +1,48 @@
...
+    return null;
+  }
+
+  @Override
+  public User getPrincipal() {
+    return principal;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
hei-teacher, line 30: use Lombok @EqualsAndHashCode

src/main/java/school/hei/haapi/authentication/JwtConfiguration.java
```diff
@@ -0,0 +1,56 @@
+package school.hei.haapi.authentication;
+
+import java.io.Serializable;
+import org.springframework.stereotype.Component;
+
+@Component
+public class JwtConfiguration implements Serializable {
+  private static final String USER_POOL_ID = ("us-east-1_AvlIh6liI");
```
hei-teacher, line 8: NOT in hard-coded like this. You should use a `final String userPoolId` which you will define in your `application-ryan.properties`. Then you launch the app with `ryan` profile

src/main/java/school/hei/haapi/authentication/JwtConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+  private static final String REGION = ("us-east-1");
```
hei-teacher, line 9: Make it configurable

src/main/java/school/hei/haapi/authentication/JwtConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+  private static final String HTTP_HEADER = "Authorization";
+  private static final int CONNECTION_TIMEOUT = 2000;
```
hei-teacher, line 11: Make it configurable

hei-teacher, line 11: Make it configurable

hei-teacher (changes requested):

src/main/java/school/hei/haapi/authentication/JwtProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JwtProcessor {
+  private static final JWSAlgorithm RS256 = new JWSAlgorithm("RS256");
+  private static final int CONNECT_TIMEOUT = 2000;
+  private static final int READ_TIMEOUT = 2000;
+  private final JwtConfiguration jwtConfiguration;
+
+  public JwtProcessor(final JwtConfiguration jwtConfiguration) {
```
hei-teacher, line 24: @AllArgsConstructor lombok

src/main/java/school/hei/haapi/authentication/JwtProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+import com.nimbusds.jwt.proc.DefaultJWTProcessor;
+import java.net.MalformedURLException;
+import java.net.URL;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JwtProcessor {
+  private static final JWSAlgorithm RS256 = new JWSAlgorithm("RS256");
+  private static final int CONNECT_TIMEOUT = 2000;
```
hei-teacher, line 20: Make it configurable

src/main/java/school/hei/haapi/authentication/JwtProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+  private static final int READ_TIMEOUT = 2000;
```
hei-teacher, line 21: Make it configurable

src/main/java/school/hei/haapi/authentication/JwtConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+  public String getJwkUrl() {
+    return this.jwkUrl != null && !this.jwkUrl.isEmpty()
+        ? this.jwkUrl
+        : String.format(
+            "https://cognito-idp.%s.amazonaws.com/%s/.well-known/jwks.json",
+            JwtConfiguration.REGION, JwtConfiguration.USER_POOL_ID);
+  }
+
+  public String getCognitoIdentityPoolUrl() {
+    return String.format("https://cognito-idp.%s.amazonaws.com/%s", JwtConfiguration.REGION, JwtConfiguration.USER_POOL_ID);
```
hei-teacher, line 26: Use constant for the String. Eg: COGNITO_USER_POOL_URL_FORMAT. And refactor with above url. Eg: COGNITO_JWKS_URL_FORMAT = COGNITO_USER_POOL_URL_FORMAT + "..."

hei-teacher (changes requested):

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-ryan: So is the role static from now ?

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-ryan: Exceptions thrown.

src/main/java/school/hei/haapi/authentication/JwtConfiguration.java
hei-ryan: Do we specify the value of the userPoolId  in the `application-profile.properties` or set the variable manually in the environment before and specify its name instead ?

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
hei-ryan: In our case I just used the email as we said.

src/main/java/school/hei/haapi/authentication/JwtConfiguration.java
hei-ryan: I was stuck here for hours.
Finally I just set environment variables in the JVM Options before the launch but even if the profile 'ryan' is active, I think I don't use properly the configuration ...

hei-ryan:
I've just found that the User Model should be renamed when I have configured the authority where the spring User and the User saved in database mismatch.
But the feat actually works.

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
```diff
@@ -0,0 +1,32 @@
...
+@Component
+public class AwsCognitoJwtAuthFilter extends GenericFilter {
+
+  private final AwsCognitoJwtIdToken cognitoIdTokenProcessor;
+
+  public AwsCognitoJwtAuthFilter(final AwsCognitoJwtIdToken cognitoIdTokenProcessor) {
+    this.cognitoIdTokenProcessor = cognitoIdTokenProcessor;
+  }
+
+  @SneakyThrows
```
hei-teacher, line 22: I used to use SneakyThrows. But now I think it's not a good idea to use it in prod. Please remove SneakyTrhows

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtIdToken.java
```diff
@@ -0,0 +1,79 @@
...
+  private String getUserNameFrom(JWTClaimsSet claims) {
+    return claims.getClaims().get(this.jwtConfiguration.getUserNameField()).toString();
+  }
+
+  private String getEmail(JWTClaimsSet claims) {
+    return claims.getClaims().get(this.jwtConfiguration.getEmailField()).toString();
+  }
+
+  private void verifyIfIdToken(JWTClaimsSet claims) throws NotAuthorizedException {
+    if (!claims.getIssuer().equals(this.jwtConfiguration.COGNITO_IDENTITY_POOL_URL_FORMAT)) {
```
hei-teacher, line 62: 1. Use Yoda conditions https://en.wikipedia.org/wiki/Yoda_conditions
2. Use static import for JwtConfiguration.COGNITO_IDENTITY_POOL_URL_FORMAT so that you'll just use COGNITO_IDENTITY_POOL_URL_FORMAT here

src/main/resources/application-ryan.properties
```diff
@@ -0,0 +1,15 @@
+spring.datasource.url=jdbc:postgresql://localhost:5433/hei_admin
```
hei-teacher, line 1: Do not persist your personal env

src/main/resources/application.properties
```diff
@@ -1 +1,6 @@
-
+spring.profiles.active=
```
hei-teacher, line 1: ndry... you don't set it here... please remove this line

src/main/resources/db/migration/V1_0__Create_User.sql
```diff
@@ -0,0 +1,11 @@
+--------------
+---- User ----
+--------------
+
+create table if not exists users (
+    id          serial constraint user_pk primary key,
+    first_name   varchar(50) not null,
+    last_name    varchar(50) not null,
+    email       varchar(50) not null constraint email_unique unique,
```
hei-teacher, line 9: add index on email and last_name as we will later search users by these keys

Explanation here https://www.facebook.com/409699499865702/videos/565820090983104

hei-teacher (changes requested):

src/main/resources/db/migration/V1_0__Create_User.sql
```diff
@@ -0,0 +1,11 @@
...
+    role        varchar(50) not null
```
hei-teacher, line 10: create a custom enum type

src/main/resources/db/migration/V1_0__Create_User.sql
```diff
@@ -0,0 +1,11 @@
+--------------
+---- User ----
+--------------
+
+create table if not exists users (
+    id          serial constraint user_pk primary key,
```
hei-teacher, line 6: not a serial, but an uuid. there already is a specific postgres module that you can activate and you'll have the uuid type. i'll leave you with the research for that

hei-teacher (changes requested):

src/main/java/school/hei/haapi/authentication/AwsCognitoJwtAuthFilter.java
hei-ryan: The method doFilter is an override method that doesn't throw the authentication Exceptions (NotAuthorizedException, ParseException, BadJOSEException, JOSEException)
So either I use sneakyThrows or i try/catch the exception.

src/test/java/school/hei/haapi/integration/conf/ContextInitializer.java
```diff
@@ -0,0 +1,40 @@
...
+        "server.port=8080",
+
+        "spring.datasource.url=" + postgreSQLContainer.getJdbcUrl(),
+        "spring.datasource.username=" + postgreSQLContainer.getUsername(),
+        "spring.datasource.password=" + postgreSQLContainer.getPassword(),
+        "spring.flyway.locations=classpath:/db/migration," + FLYWAY_TESTDATA_PATH,
+
+        "reverse.admin.token=admin-token",
+        "reverse.monitor.token=monitor-token",
+        "reverse.pagination.limit.max=" + PAGINATION_LIMIT_MAX,
```
hei-teacher, line 36: remove these reverse params

hei-teacher (changes requested):

src/main/resources/db/migration/V1_0__Create_User.sql
```diff
@@ -0,0 +1,15 @@
+--------------
+---- User ----
+--------------
+create type "role" as enum ('STUDENT','TEACHER','STAFF');
```
hei-teacher, line 4: STAFF --> MANAGER

Also, check whether the type exists before its creation. either `create type if  not exists...` (or try `drop type if exists...` before creation). The goal is to be able to run the script multiple times but with the same ultimate effect. This is called idempotency

src/main/resources/db/migration/V1_0__Create_User.sql
```diff
@@ -0,0 +1,15 @@
...
+
+create table if not exists "user" (
+    id          varchar constraint user_pk primary key,
+    first_name   varchar not null,
+    last_name    varchar not null,
+    email       varchar not null constraint email_unique unique,
+    role        role not null
+);
+
+create index on "user" (email);
```
hei-teacher, line 14: if not exists

src/main/resources/db/migration/V1_0__Create_User.sql
```diff
@@ -0,0 +1,15 @@
...
+create index on "user" (last_name);
```
hei-teacher, line 15: if not exists

hei-teacher (changes requested):

src/main/resources/db/migration/V1_0__Create_User.sql
hei-ryan: But for what purpose would we run the script multiple times if we want to have the same effect ?

src/test/java/school/hei/haapi/integration/SecurityIT.java
```diff
@@ -0,0 +1,109 @@
...
+  public SecurityIT(@LocalServerPort int serverPort) {
+    this.serverPort = serverPort;
+  }
+
+  private HAHttpClient clientFromToken(String token) {
+    return new HAHttpClient("http://localhost:" + serverPort, token);
+  }
+
+  @Test
+  @SneakyThrows
```
hei-teacher, line 37: Remove

src/test/java/school/hei/haapi/integration/SecurityIT.java
```diff
@@ -0,0 +1,109 @@
...
+   * UNAUTHORIZED/NOT_AUTHORIZED and FORBIDDEN, _but_ to pass the test I modified it to
+   * UNAUTHORIZED, so you can easily do the test while waiting your comments.
+   */
+
+  @Test
+  void student1_can_not_get_student2_information() {
+    // given
+    HAHttpClient student1 = clientFromToken(CallerData.STUDENT1_TOKEN);
+
+    // excepted
```
hei-teacher, line 61: remove the given-when-expected comments. Just put them on a test for illustration and do not put them for the rest

src/main/java/school/hei/haapi/web/rest/UserController.java
```diff
@@ -0,0 +1,50 @@
...
+  private final UserService userService;
+  private final UserMapper userMapper;
+
+  @GetMapping(value = "/users/{userId}")
+  public StudentResource getUserById(
+      @AuthenticationPrincipal ApiClient client, @PathVariable String userId) {
+    User user = userService.getById(userId);
+    String clientRole = client.getRole();
+    String userRole = user.getRole();
+    // Student exceptions
```
hei-teacher, line 29: remove all these comments here. The code is already explicit enough

src/test/java/school/hei/haapi/integration/SecurityIT.java
```diff
@@ -0,0 +1,109 @@
...
+import school.hei.haapi.web.model.StudentResource;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT;
+
+@SpringBootTest(webEnvironment = RANDOM_PORT)
+@Testcontainers
+@ContextConfiguration(initializers = ContextInitializer.class)
+@AutoConfigureMockMvc
```
hei-teacher, line 23: If you don't use mock anymore, remove this.

However I suggest you continue mocking Cognito responses here. Then separately test cognitoComponent in another unit test where you only check the cognitoComponent::findByEmail.

This way, the authorization tests will always work even if the authentication tests fail (eg if tokens expire)

src/test/java/school/hei/haapi/integration/conf/CallerData.java
```diff
@@ -0,0 +1,14 @@
...
+  // This token expires 2021/08/25 16:00
+  public static final String STUDENT1_TOKEN =
+      "eyJraWQiOiJPam5ZXC9YVW9FVzlNVkwzbWs2bTR1a2xqMHFOdUE0WXdRRlpTbzd2MGNUOD0iLCJhbGciOiJSUzI1NiJ9.eyJhdF9oYXNoIjoiaTg4YTNHVWhaWGsxZmlGV2JaTWZPUSIsInN1YiI6IjZmYWY4ZjllLWQ2ZDQtNDE0Zi05MjE3LTMzZWEyZGRiNjczNCIsImVtYWlsX3ZlcmlmaWVkIjp0cnVlLCJpc3MiOiJodHRwczpcL1wvY29nbml0by1pZHAudXMtZWFzdC0yLmFtYXpvbmF3cy5jb21cL3VzLWVhc3QtMl9nS2RZM1JoMFgiLCJwaG9uZV9udW1iZXJfdmVyaWZpZWQiOnRydWUsImNvZ25pdG86dXNlcm5hbWUiOiJyeWFuIiwiYXVkIjoiMzd0aWhnNGZmaDU3ZGp2NDNqcDNoNDd2bWYiLCJldmVudF9pZCI6IjE2ZjhmNjhjLTA2YzctNDVjNS1hYjZjLTUyYjU5ZTVjZmFjNiIsInRva2VuX3VzZSI6ImlkIiwiYXV0aF90aW1lIjoxNjI5ODEzMTEzLCJwaG9uZV9udW1iZXIiOiIrMjYxMzQxMjM0NTY3IiwiZXhwIjoxNjI5ODk5NTEzLCJpYXQiOjE2Mjk4MTMxMTMsImp0aSI6Ijc3ZGE2ZTQyLTRmNzQtNGMyMS05ZTU0LTNkZWU0ZTFlNWJiZCIsImVtYWlsIjoicnlhbkBoZWkuc2Nob29sIn0.ehnhU47AtNY1-PTckT8nl9FWkOkyqbb2HjiciWX2HZih8NCgGoEjIarOCJWZOJ_6EiTOEvh1gWG0umeOK6z8NJD2vEPZ_fO-SkptdOsrExgtut5bzhvDdcJ4TY4MfhPqkbrFmRKQr76joE_IniudErWBglU8rv6ILTS95WufXygXyBLZ2f-Slbm-ZcaCaJ22EPzGifr62rfyfa2QMB2oq-dg0L9V7bcP7Qdaj4_-0s_-uaG3cTm5_v0WzGuTtp2EH0LB2a5YD4FCWaEYLeQl0_9uF20aYnDoOc3lE7rFbmSWKUFcZYPrQUDNutHvAiUJPC23olBHrLgWpGtOA83afg";
```
hei-teacher, line 10: So that it is clear once and for all: never EVER EVER EVER put authentication token in public read like this!!!!!

Put them in env variables.

We will eventually use [localstack](https://github.com/localstack/localstack) to emulate Cognito locally, but that's in a distant future. For now, just use direct call to  online Cognito as you already does, but separate its tests from the authorization test.

hei-teacher (changes requested):

src/test/java/school/hei/haapi/integration/conf/CallerData.java
hei-ryan: the "its" in "its tests" refer to what ? 
I don't really understand what should I separate from the authorization test as our tests  are _all_ based on authorization.

src/test/java/school/hei/haapi/integration/SecurityIT.java
hei-ryan: So you suggest me to create an duplicated IT but in one hand using the mocks and in another hand just to get the result ?

src/main/resources/db/migration/V1_0__Create_User.sql
hei-teacher: https://www.red-gate.com/hub/product-learning/flyway/creating-idempotent-ddl-scripts-for-database-migrations#:~:text=The%20solution%3A%20idempotent%20SQL%20migration%20scripts&text=It%20is%20really%20a%20fancy,safely%20into%20a%20Flyway%20migration.

src/test/java/school/hei/haapi/integration/conf/CallerData.java
hei-teacher: **Test class 1 : Authentication**, with direct call to Cognito. You don't test the API. You don't need Postgres. You just test the CognitoComponent class.

**Test class 2 : Authorization**, with mock Cognito. You suppose that CoginoComponent is working properly and now you the API and the DB.

it.patch
```diff
@@ -0,0 +1,43 @@
+ build.gradle                                       |   3 +
```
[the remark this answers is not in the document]
hei-teacher: do not resolve conversations until their resolution are pushed

src/main/java/school/hei/haapi/security/model/Role.java
```diff
@@ -0,0 +1,18 @@
+package school.hei.haapi.security.model;
+
+import org.springframework.security.core.GrantedAuthority;
+
+public enum Role implements GrantedAuthority {
+  STUDENT,
+  TEACHER,
+  STAFF;
```
hei-teacher, line 8: MANAGER. Already commented before

hei-teacher (changes requested):

hei-teacher (changes requested):

src/main/resources/application.properties
```diff
@@ -1 +1,20 @@
...
+#test.cognito.jwt.studenttoken=
+cognito.jwt.jwsAlgorithm=RS256
+cognito.jwt.readTimeout=2000
+cognito.jwt.connectTimeout=2000
 
+server.port=8080
+#spring.datasource.url=jdbc:postgresql:hei_admin
+#spring.datasource.username=postgres
+#spring.datasource.password=postgres
+#spring.jpa.hibernate.ddl-auto=none
```
hei-teacher, line 12: do not comment

src/main/resources/application.properties
```diff
@@ -1 +1,20 @@
...
+#spring.jpa.show-sql=false
```
hei-teacher, line 13: do not comment

src/main/resources/application.properties
```diff
@@ -1 +1,20 @@
...
+
+#spring.flyway.user=postgres
+#spring.flyway.password=postgres
+#spring.flyway.url=jdbc:postgresql:hei_admin
+#spring.flyway.locations=classpath:/db/migration,classpath:/db/testdata
```
hei-teacher, line 18: remove lasspath:/db/testdata then do not comment

hei-teacher (changes requested):

src/main/resources/application.properties
```diff
@@ -1 +1,20 @@
+#cognito.jwt.userPoolId=
+#cognito.jwt.region=us-east-2
+#test.cognito.jwt.studenttoken=
+cognito.jwt.jwsAlgorithm=RS256
+cognito.jwt.readTimeout=2000
+cognito.jwt.connectTimeout=2000
 
+server.port=8080
+#spring.datasource.url=jdbc:postgresql:hei_admin
+#spring.datasource.username=postgres
```
hei-teacher, line 10: do not comment

hei-teacher (changes requested):

src/main/resources/application.properties
```diff
@@ -1 +1,20 @@
...
+#spring.datasource.password=postgres
+#spring.jpa.hibernate.ddl-auto=none
+#spring.jpa.show-sql=false
+
+#spring.flyway.user=postgres
```
hei-teacher, line 15: ne pas commenter

hei-teacher (changes requested):

Issue #10, opened 2021-12-05
Locally run Cognito during tests


hei-teacher:
Commit https://github.com/hei-school/hei-admin-api/commit/74166d4145451f3086804bb23d3f171485b44355 uses the Cognito emulator from https://github.com/jagregory/cognito-local. Unfortunately, that emulator does *not* support `AdminInitiateAuth` which makes it impossible to get authentication token. Keep an eye on that project, they might eventually implement that feature.

Alternately, Localstack provides support for Cognito. However, they do so in the Pro version only https://localstack.cloud/features/.

PR #11, opened 2021-12-05
Disable test against real Cognito & rework packages hierarchy

hei-teacher:
Disable testing authentication against online Cognito while waiting for its local implementation, as described in  https://github.com/hei-school/hei-admin-api/issues/10.


src/main/resources/application.properties
```diff
@@ -1,8 +1,6 @@
 server.port=8080
 
-#cognito.jwt.userPoolId=
-#cognito.jwt.region=us-east-2
-#test.cognito.jwt.studenttoken=
+#cognito.userPool.url=https://cognito-idp.<region>.amazonaws.com/<userPoolId>
```
hei-teacher, line 5: @hei-ryan Set `secrets.COGNITO_USERPOOL_URL` before merging.

PR #34, opened 2022-01-17
feat: GET /fees/{id}

hei-teacher:


doc/api.yml
```diff
@@ -793,6 +793,29 @@ components:
             - UNPAID
             - PAID
             - LATE
+        comments:
+          type: array
+          items:
+            $ref: '#/components/schemas/Comment'
+    Comment:
+      allOf:
+        - $ref: '#/components/schemas/CreateComment'
```
hei-teacher: There is no need for the CreateComment resource. Comment is part of Fee. You create comment by creating Fee with comment. That POST /fees where, the comment field is set.

PR #42, opened 2022-01-28
test: userUpsert events

hei-ryan:


doc/api.yml
```diff
@@ -563,16 +563,19 @@ paths:
...
       tags:
         - Paying
       summary: Create student fees
-      parameters:
-        - name: student_id
-          in: path
-          required: true
-          schema:
-            type: string
+      requestBody:
```
hei-teacher: This modification should be done in a separate commit as it has nothing to do with the fix we are trying to do. Next time, use ONE commit for doing ONE thing. This time, I will let it pass, BUT give an adequate name first.

`test: userUpser events` does NOT reflect the main purpose of this commit...

src/main/java/school/hei/haapi/endpoint/event/EventConf.java
```diff
@@ -9,6 +9,7 @@
 @Configuration
 public class EventConf {
   private final Region region;
+  public final static Integer MAX_EVENT_REQUEST_ENTRY = 10;
```
hei-teacher: Why `Integer` instead of `int`?

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
```diff
@@ -41,19 +44,18 @@ public Student getStudentById(
...
+    return userService.getByRole(User.Role.STUDENT, page, pageSize).stream()
         .map(userMapper::toRestStudent)
         .collect(toUnmodifiableList());
   }
 
   @PutMapping("/students")
-  public List<Student> saveAll(@RequestBody List<Student> toSave) {
+  public List<Student> saveAll(@RequestBody List<Student> toWrite) {
+    if (toWrite.size() > EventConf.MAX_EVENT_REQUEST_ENTRY) {
+      throw new BadRequestException("students to create or update must be ≤ 10");
```
hei-teacher: Our exception message starts with a capital letter. Also, use MAX_EVENT_REQUEST_ENTRY instead of the magic number 10

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
```diff
@@ -41,19 +44,18 @@ public Student getStudentById(
...
+    }
     return userService
-        .saveAll(toSave
-            .stream()
-            .map(userMapper::toDomain)
-            .collect(toUnmodifiableList()))
+        .saveAll(toWrite.stream().map(userMapper::toDomain).collect(toUnmodifiableList()))
```
hei-teacher, line 52: I prefer previous indentation

src/main/java/school/hei/haapi/endpoint/rest/controller/TeacherController.java
```diff
@@ -40,19 +42,19 @@ public Teacher getTeacherById(
...
-    var saved = userService.saveAll(toWrite.stream()
-        .map(userMapper::toDomain)
-        .collect(toUnmodifiableList()));
-    return saved.stream()
-        .map(userMapper::toRestTeacher)
-        .collect(toUnmodifiableList());
+    if (toWrite.size() > EventConf.MAX_EVENT_REQUEST_ENTRY) {
+      throw new BadRequestException("teachers to create or update must be ≤ 10");
+    }
+    var saved =
```
hei-teacher: Do not modify indentation unless it's really really bad. It just makes reviewing the PR harder

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -17,13 +19,14 @@
 import static java.util.stream.Collectors.toUnmodifiableList;
 import static org.springframework.data.domain.Sort.Direction.ASC;
 
+@Slf4j
+
```
hei-teacher: NO

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -17,13 +19,14 @@
...
 @Service
 @AllArgsConstructor
 public class UserService {
 
   private final UserRepository userRepository;
   private final EventProducer eventProducer;
-
```
hei-teacher: NO

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -32,11 +35,13 @@ public User getByEmail(String email) {
     return userRepository.getByEmail(email);
   }
 
+  @Transactional
   public List<User> saveAll(List<User> users) {
     List<User> savedUsers = userRepository.saveAll(users);
     eventProducer.accept(users.stream()
         .map(this::toTypedEvent)
         .collect(toUnmodifiableList()));
+
```
hei-teacher: NO

src/test/java/school/hei/haapi/integration/PaginationIT.java
```diff
@@ -67,21 +64,22 @@ private void createNewStudents(int nbOfNewStudents) throws ApiException {
...
-    List<Student> page100 = api.getStudents(100, 5);
 
-    assertEquals(5, page5.size());
-    assertEquals(5, page10.size());
+    List<Student> page1 = api.getStudents(1, pageSize);
+    List<Student> page2 = api.getStudents(2, pageSize);
+    List<Student> page100 = api.getStudents(100, pageSize);
+
+    assertEquals(pageSize, page1.size());
+    assertEquals(pageSize, page2.size());
```
hei-teacher, line 78: Also get page3 and tests that its size equals 2

src/test/java/school/hei/haapi/integration/StudentIT.java
```diff
@@ -77,16 +74,22 @@ public static Student aCreatableStudent() {
...
-        .toInstant();
+    Instant birthday = faker.date().birthday().toInstant();
     int ageOfEntrance = 14 + (int) (Math.random() * 20);
     student.setBirthDate(birthday.atZone(ZoneId.systemDefault()).toLocalDate());
     student.setEntranceDatetime(birthday.plus(365L * ageOfEntrance, ChronoUnit.DAYS));
     student.setAddress(faker.address().fullAddress());
     return student;
   }
 
+  static List<Student> aStudentList(int number) {
```
hei-teacher: aCreatableStudentList...

src/test/java/school/hei/haapi/integration/StudentIT.java
```diff
@@ -215,21 +213,55 @@ void manager_write_update_ok() throws ApiException {
...
-        PutEventsResponse.builder()
-            .entries(
-                PutEventsResultEntry.builder().eventId("eventId1").build(),
-                PutEventsResultEntry.builder().eventId("eventId2").build())
-            .build());
-
-    List<Student> created = api.createOrUpdateStudents(List.of(
-        aCreatableStudent(),
-        aCreatableStudent()));
+    when(eventBridgeClientMock.putEvents((PutEventsRequest) any()))
```
hei-teacher: Useless indendation change that only harden my reading. Rollback to old intendation

src/test/java/school/hei/haapi/integration/TeacherIT.java
```diff
@@ -101,6 +105,14 @@ public static Teacher aCreatableTeacher() {
     return teacher;
   }
 
+  static List<Teacher> aTeacherList(int number) {
```
hei-teacher: aCreatableStudentList...

hei-teacher (changes requested):

doc/api.yml
hei-ryan: For sure !
I have already created a new branch from origin dev, but I forgot to remove the api.yml ... my bad !

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
hei-ryan: it's google_check format so ?

src/main/java/school/hei/haapi/endpoint/rest/controller/TeacherController.java
hei-ryan: still google_check format

src/main/java/school/hei/haapi/endpoint/event/EventConf.java
hei-ryan: Habits !

src/test/java/school/hei/haapi/integration/StudentIT.java
hei-ryan: Done

src/test/java/school/hei/haapi/integration/TeacherIT.java
hei-ryan: Done

src/test/java/school/hei/haapi/integration/PaginationIT.java
hei-ryan: Done

hei-ryan: Actually, page3 size equals 4 not 2 ...

hei-ryan: It was because we have already created 2 students in the database

src/test/java/school/hei/haapi/integration/StudentIT.java
hei-ryan: Done

src/main/java/school/hei/haapi/endpoint/event/EventConf.java
```diff
@@ -9,6 +9,7 @@
 @Configuration
 public class EventConf {
   private final Region region;
+  public final static int MAX_EVENT_REQUEST_ENTRY = 10;
```
hei-teacher: MAX_PUT_EVENT_ENTRIES

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
```diff
@@ -41,19 +43,21 @@ public Student getStudentById(
   @GetMapping("/students")
   public List<Student> getStudents(
       @RequestParam PageFromOne page, @RequestParam("page_size") BoundedPageSize pageSize) {
-    return userService
-        .getByRole(User.Role.STUDENT, page, pageSize).stream()
-        .map(userMapper::toRestStudent)
-        .collect(toUnmodifiableList());
+    List<Student> students =  userService.getByRole(User.Role.STUDENT, page, pageSize).stream()
```
hei-teacher: No added value in this change, keep previous code.

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
```diff
@@ -41,19 +43,21 @@ public Student getStudentById(
...
+            .map(userMapper::toRestStudent)
+            .collect(toUnmodifiableList());
+    return students;
   }
 
   @PutMapping("/students")
-  public List<Student> saveAll(@RequestBody List<Student> toSave) {
+  public List<Student> saveAll(@RequestBody List<Student> toWrite) {
+    if (toWrite.size() > EventConf.MAX_EVENT_REQUEST_ENTRY) {
```
hei-teacher: Now that I think about it, the check should probably be performed at the service level instead of the controller level. The main advantage being that we will be sure any call to `UserService.saveAll` would be checked before sending event.

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
```diff
@@ -41,19 +43,21 @@ public Student getStudentById(
...
+      throw new BadRequestException(
+          String.format(
+              "Students to create or update must be ≤ %d", EventConf.MAX_EVENT_REQUEST_ENTRY));
```
hei-teacher: Just `BadRequestException("string" + MAX_PUT_EVENT_ENTRIES)` should be enough. The use of `String.format` is just cumbersome without adding any real value.

hei-teacher (changes requested):

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
hei-teacher: Now that I think even more about it (yes, I do think a lot), the checking whether a payload to produce over the event bus is valid MUST be a responsibility of `EventProducer`, as the name suggests.

Hence you should not leak `EventConf.MAX_EVENT_REQUEST_ENTRY` publicly. You should create a **public** method `boolean EventProducer.isPayloadValid(List<T>)` which will do its calculation using a **private** `static final in EventProducer.MAX_PUT_EVENT_ENTRIES`.

Then in `UserService` you would just call `EventProducer.isPayloadValid` before calling `EventProducer.sendEvent`.

src/main/java/school/hei/haapi/service/UserService.java
hei-teacher: Do not rm...

src/main/java/school/hei/haapi/endpoint/event/EventProducer.java
```diff
@@ -25,11 +26,12 @@
 
 @Component
 @Slf4j
-public class EventProducer implements Consumer<List<TypedEvent>> {
+public class EventProducer<T> implements Consumer<List<TypedEvent>> {
```
hei-teacher: Why do you need to generify EventProducer? In particular, you never use T in your implementation so it brings really nothing

hei-teacher (changes requested):

src/main/java/school/hei/haapi/endpoint/event/EventProducer.java
```diff
@@ -108,4 +114,12 @@ private void checkResponse(PutEventsResponse response) {
       log.info("Following events were successfully sent: {}", successfulEntries);
     }
   }
+
+  public boolean isPayloadValid(List<TypedEvent> events) {
```
hei-teacher: `isXxx` is not supposed to throw an exception. It should only return either `true` or `false`.

If you want to do it through an exception thrown from `EventProducer` (which is something I like), use a `private void checkPayload(List)` instead. It will throw BadRequest in case of invalidity, and nothing in case of validity. Also, as the message will be ultimately shown to consumer, use a more generic error message (as consumer does not need to know that we are using events behind). Eg something like `Request entries must be <= ...`

hei-teacher (changes requested):

PR #105, opened 2023-01-12
chore: filter by criteria changed

Sanmandresy:
Changed the filter by criteria method that was too long


hei-ryan:
Check the conventional commit here :  https://www.conventionalcommits.org/en/v1.0.0/

PR #115, opened 2023-02-01
chore: persist remaining amount and status

Sanmandresy:


src/main/java/school/hei/haapi/service/FeeService.java
```diff
@@ -28,108 +27,54 @@ public class FeeService {
...
-        page.getValue() - 1,
-        pageSize.getValue(),
-        Sort.by(DESC, "dueDatetime"));
     if (status != null) {
-      List<Fee> studentFees = resetPaymentRelatedInfo(feeRepository.getByStudentId(studentId));
-      return getFeesByStatus(studentFees, status, page, pageSize);
+      return feeRepository.getFeesByStatus(status, pageable);
     }
-    return resetPaymentRelatedInfo(feeRepository.getByStudentId(studentId, pageable));
+    return feeRepository.getFeesByStatus(LATE, pageable);
```
hei-ryan: Set LATE in a constant name `DEFAULT_STATUS` because it was not very explicit.

src/main/java/school/hei/haapi/service/FeeService.java
```diff
@@ -28,108 +27,54 @@ public class FeeService {
...
-    if (payments != null) {
-      int amount = payments
-          .stream()
-          .mapToInt(Payment::getAmount)
-          .sum();
-      return fee.getTotalAmount() - amount;
+  public Fee updateFeeStatus(Fee initialFee) {
+    if (initialFee.getRemainingAmount() == 0) {
+      initialFee.setStatus(PAID);
+    } else if (Instant.now().isAfter(initialFee.getDueDatetime())) {
```
hei-ryan, line 71: Pay attention ! `Instant.now()` is not taking time zone so later you must set the default time zone to UTC+3.

src/main/java/school/hei/haapi/service/FeeService.java
```diff
@@ -28,108 +27,54 @@ public class FeeService {
...
 
-  private List<Fee> resetPaymentRelatedInfo(List<Fee> fees) {
-    return fees.stream()
-        .map(this::resetPaymentRelatedInfo)
-        .collect(toUnmodifiableList());
+  @Scheduled(cron = "0 0 * * * *")
+  public void updateFeesStatusToLate() {
+    List<Fee> unpaidFees = feeRepository.getUnpaidFees();
+    unpaidFees.forEach(this::updateFeeStatus);
+    feeRepository.saveAll(unpaidFees);
```
hei-ryan, line 84: Log each time an action is done in a scheduler.

For example `log.info("Fee.id was updated from UNPAID to LATE")`

src/main/java/school/hei/haapi/service/PaymentService.java
```diff
@@ -8,32 +8,45 @@
...
+                                              BoundedPageSize pageSize) {
+    Pageable pageable = PageRequest.of(page.getValue() - 1,
         pageSize.getValue(),
         Sort.by(DESC, "creationDatetime"));
     return paymentRepository.getByStudentIdAndFeeId(studentId, feeId, pageable);
   }
 
+  public void computeRemainingAmount(String feeId, int amount) {
+    Fee associatedFee = feeService.getById(feeId);
+    int computedAmount = associatedFee.getRemainingAmount() - amount;
```
hei-ryan: Remove this `computedAmount` variable and use directly its value.

src/main/java/school/hei/haapi/service/PaymentService.java
```diff
@@ -8,32 +8,45 @@
...
+    associatedFee.setRemainingAmount(computedAmount);
+    if (associatedFee.getRemainingAmount() == 0) {
+      associatedFee.setStatus(PAID);
+    }
+  }
+
   @Transactional
```
hei-ryan: Use @Transactional(isolation=SERIALIZABLE)

src/main/resources/application.properties
```diff
@@ -1,26 +1,28 @@
 server.port=8080
 
-#env=
+#env=dev
```
hei-ryan: Remove `dev`

src/main/resources/application.properties
```diff
@@ -1,26 +1,28 @@
...
 
-#spring.datasource.url=jdbc:postgresql:hei_admin
+#spring.datasource.url=jdbc:postgresql:/hei_admin
 #spring.datasource.password=postgres
 spring.datasource.username=postgres
 spring.flyway.locations=classpath:/db/migration
 spring.jpa.hibernate.ddl-auto=none
 spring.jpa.show-sql=false
 spring.jpa.open-in-view=true
+spring.jpa.properties.hibernate.jdbc.time_zone=UTC+3
```
hei-ryan: If you set time zone as UTC+3 in database, note that Java Instant is not configured through this.

src/main/resources/db/migration/V0_7__Update_fee_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    return new;
+end;
+$$ language 'plpgsql';
+
+create trigger update_status_updated_time
+    before update
+    on
+        "fee"
+    for each row
+execute procedure update_status_updatedtime();
```
hei-ryan: So you mean, if I update the comment of the fee, the `status_updated_time` is updated ? 

Nor you should rename the status_updated_time to `updatedAt`

Nor you should not update the status_updated_time for EVERY update, only when update is set on fee_status.

hei-ryan (changes requested):

src/main/java/school/hei/haapi/HaApiApplication.java
```diff
@@ -10,4 +12,9 @@ public class HaApiApplication {
   public static void main(String[] args) {
     SpringApplication.run(HaApiApplication.class, args);
   }
+
+  @PostConstruct
+  public void init() {
+    TimeZone.setDefault(TimeZone.getTimeZone("UTC+3"));
```
hei-ryan: Have you created test for this ?

src/main/java/school/hei/haapi/repository/FeeRepository.java
```diff
@@ -3,14 +3,23 @@
...
 
+  List<Fee> getFeesByStatus(StatusEnum status, Pageable pageable);
+
+  List<Fee> getFeesByStudentIdAndStatus(String studentId, StatusEnum status, Pageable pageable);
+
   List<Fee> getByStudentId(String studentId, Pageable pageable);
 
-  List<Fee> getByStudentId(String studentId);
+  @Query("select f from Fee f where f.status = 'UNPAID' "
+      + "and f.remainingAmount != 0 "
```
hei-ryan: Use the operator > instead of != to be more explicit.

src/main/java/school/hei/haapi/service/FeeService.java
```diff
@@ -1,135 +1,86 @@
...
 
-  private int computeRemainingAmount(Fee fee) {
-    List<Payment> payments = fee.getPayments();
-    if (payments != null) {
-      int amount = payments
-          .stream()
-          .mapToInt(Payment::getAmount)
-          .sum();
-      return fee.getTotalAmount() - amount;
+  public Fee updateFeeStatus(Fee initialFee) {
```
hei-ryan, line 68: Can't you just put this method into the mapper so it's generalized to avoid to call it each time ?

For example during toEntity if it should be persisted after or toDomain if it's calculated.

hei-ryan (changes requested):

src/main/java/school/hei/haapi/service/FeeService.java
Sanmandresy: But this method is used in the Scheduler actually...

src/main/java/school/hei/haapi/service/FeeService.java
```diff
@@ -1,135 +1,86 @@
...
-  private List<Fee> resetPaymentRelatedInfo(List<Fee> fees) {
-    return fees.stream()
-        .map(this::resetPaymentRelatedInfo)
-        .collect(toUnmodifiableList());
+  @Scheduled(cron = "0 0 * * * *")
+  public void updateFeesStatusToLate() {
+    List<Fee> unpaidFees = feeRepository.getUnpaidFees();
+    unpaidFees.forEach(fee -> {
+      updateFeeStatus(fee);
+      log.info("Fee with id." + fee.getId() + " was updated from UNPAID to LATE");
```
hei-ryan: Because this method is before the update, reformat the message as.

was udpated -> is going to be updated

src/test/java/school/hei/haapi/integration/SpringSecurityIT.java
```diff
@@ -106,6 +109,14 @@ void test_cors(HttpMethod method, String path) throws IOException, InterruptedEx
     assertEquals("authorization", headersList.get(0));
   }
 
+  @Test
```
hei-ryan, line 113: Add a TODO, this test should not be here but for now we can leave it here.

src/test/java/school/hei/haapi/integration/SpringSecurityIT.java
```diff
@@ -106,6 +109,14 @@ void test_cors(HttpMethod method, String path) throws IOException, InterruptedEx
...
+  void check_timezone_is_utc_plus_three() {
+    ZoneId zoneId = ZoneId.of("Indian/Antananarivo");
+    String utcPlusThree = "+03:00";
```
hei-ryan: utcPlusThree -> utc

hei-ryan (changes requested):

PR #273, opened 2023-10-05
Feat/attendance

haritianaadriano:


haritianaadriano:
The link to the sequence diagram is attached: [The link to the sequence diagram is attached:](diagramme de sequence)

PR #350, opened 2024-01-05
docs(api): add fee types endpoint

Iloniavo:


doc/api.yml
```diff
@@ -1565,6 +1565,83 @@ paths:
...
+      operationId: createOrUpdateFeeType
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/FeeType'
+      responses:
+        '200':
+          description: The created or updated fee type
```
Mahefaa, line 1636: plural

Mahefaa: if you are creating a FeeType then add a /id to the path for the put.

If you intend to create multiple types, then set a List as requestBody

Mahefaa (changes requested):

doc/api.yml
```diff
@@ -2304,6 +2388,20 @@ components:
...
+          type: string
+        name:
+          type: string
+        total_amount:
+          type: integer
+        number_of_months:
+          type: integer
+        creation_datetime:
+          type : string
+          format : date-time
```
Mahefaa: separate into two entities, one CreateFeeType without the creation_datetime and inherit the properties of CreateFeeType for FeeType

Mahefaa (changes requested):

Mahefaa (approved):

PR #351, opened 2024-01-05
fix: use common core as default value for student

haritianaadriano:


src/main/resources/db/migration/V42_8__Update_user_table.sql
```diff
@@ -0,0 +1,2 @@
+alter table "user" drop column specialization_field;
+alter table "user" add column if not exists specialization_field specialization_field default 'COMMON_CORE';
```
Mahefaa: use alter column and set default

Mahefaa (changes requested):

Mahefaa (approved):

PR #353, opened 2024-01-08
fix: default student specialization field to common core

haritianaadriano:


src/main/resources/db/migration/V42_9__Update_student_data.sql
```diff
@@ -0,0 +1,2 @@
+update "user" u set specialization_field = 'COMMON_CORE'
+    where u.role = 'STUDENT';
```
Mahefaa: set it to common_core where it is null

Mahefaa (changes requested):

Mahefaa (approved):

PR #361, opened 2024-01-17
docs(api): add list of feeTypeComponent in fee type model

Iloniavo:


doc/api.yml
```diff
@@ -2429,6 +2429,19 @@ components:
           type: integer
         number_of_months:
           type: integer
+        fees:
+          type: array
+          items:
+            $ref: '#/components/schemas/FeeTypeComponent'
```
Mahefaa, line 2435: not fees

doc/api.yml
```diff
@@ -2429,6 +2429,19 @@ components:
...
+    FeeTypeComponent:
+      type: object
+      properties:
+        id:
+          type: string
+        month_number:
+          type: integer
```
Mahefaa, line 2442: add description.

i think it is even better if you add an enum for months.
From January to December

Mahefaa (changes requested):

PR #384, opened 2024-02-08
docs(api): upload transfer receipt

Iloniavo:


doc/api.yml
```diff
@@ -2473,6 +2514,19 @@ components:
...
+      type: object
+      properties:
+        id:
+          type: string
+        student_id:
+          type: string
+        fee_id:
+          type: string
+        type:
+          $ref: '#/components/schemas/FeeTypeEnum'
```
Mahefaa, line 2532: how is this guessed from the image ?

Mahefaa (changes requested):

doc/api.yml
Iloniavo: Rename to s3_image_url ?

PR #388, opened 2024-02-15
feat: add longitude and latitude attributes to user

Iloniavo:


Mahefaa (changes requested):
put longitude and latitude in a different property name coordinates.
add a validator for these coordinates ( neither long nor lat is nullable).

Mahefaa (approved):

PR #402, opened 2024-03-04
feat: inject event gen class dependency with all args

haritianaadriano:


src/main/java/school/hei/haapi/endpoint/rest/controller/FeeController.java
```diff
@@ -38,13 +38,6 @@ public class FeeController {
...
   private final FeeTemplateMapper feeTemplateMapper;
 
-  @GetMapping("/fees/late")
-  public List<Fee> getLateFee() {
-    return feeService.getLateFees().stream()
-        .map(feeMapper::toRestFee)
-        .collect(toUnmodifiableList());
-  }
-
   @PutMapping("/fees/late")
```
Mahefaa: do you need to keep this ??
I'm pretty sure i have told you to revert the commits once done testing !

Mahefaa (changes requested):

PR #409, opened 2024-03-08
fix: picture url removed after updating user

haritianaadriano:


doc/api.yml
```diff
@@ -2723,6 +2723,8 @@ components:
             entrance_datetime:
               type: string
               format: date-time
+            specialization_field:
+              $ref: '#/components/schemas/SpecializationField'
```
Mahefaa: rm specialization_field from Student if you put it in createUser, also, doesn't it mean that all Users have SpecializationField now ?

Mahefaa (changes requested):

PR #422, opened 2024-04-02
docs(api): announcements endpoint

Iloniavo:


doc/api.yml
```diff
@@ -3466,6 +3643,58 @@ components:
...
+          type: array
+          items:
+            type: string
+          example: ["G1", "G2"]
+    Scope:
+      type: string
+      enum:
+        - GLOBAL
+        - TEACHER
+        - STUDENT
```
Mahefaa, line 3672: add MANAGER

doc/api.yml
```diff
@@ -3466,6 +3643,58 @@ components:
...
+    Announcement:
+      type: object
+      properties:
```
Mahefaa, line 3676: you forgot the scope ?

Mahefaa (changes requested):

PR #423, opened 2024-04-02
feat!: migrate db to neon

Dinasoa:


src/main/resources/db/migration/V42_2__Create_dummy_uuid_table.sql
```diff
@@ -6,4 +6,4 @@ create table if not exists dummy_uuid
 
 insert into dummy_uuid (id)
 values ('dummy-uuid-id-1')
-on conflict on constraint dummy_uuid_pk do nothing;
```
Mahefaa, line 9: why did you do that ?
it will raise conflicts with existing history

src/main/resources/db/migration/V42_1__Create_dummy_table.sql
```diff
@@ -6,4 +6,4 @@ create table if not exists dummy
 
 insert into dummy (id)
 values ('dummy-table-id-1')
-on conflict on constraint dummy_pk do nothing;
```
Mahefaa, line 9: why did you do that ?
it will raise conflicts with existing history

put a breaking change if it is really necessary

Mahefaa (changes requested):

PR #448, opened 2024-04-23
refactor: SecurityConf to retrieve user on permit all request

haritianaadriano:


Mahefaa (changes requested):
bearerFilter has to contain the requestMatchers requiring full authentication

PR #456, opened 2024-05-03
fix: get fees by id then update status to paid or late

haritianaadriano:


Mahefaa (approved):
approved,has this change been applied yet ?

Mahefaa:
merged differently as https://github.com/hei-school/hei-admin-api/blob/a1457e89662ea0350c07d85f4ff1b904551aa111/src/main/java/school/hei/haapi/service/FeeService.java#L101-L129

PR #487, opened 2024-06-13
refactor: student status not required when defining student status

haritianaadriano:


Mahefaa (changes requested):
where is the attribute removal ?
# my bad it's present !

Mahefaa (approved):

PR #488, opened 2024-06-13
fix: students stat are now exact

Iloniavo:


doc/api.yml
```diff
@@ -4260,6 +4256,36 @@ components:
...
+            $ref: '#/components/schemas/Student'
+    GroupStat:
+      type: object
+      properties:
+        total:
+          type: integer
+        men:
+          type: integer
+        women:
+          type: integer
```
Mahefaa: looks duplicated.
you simply have to add students "en alternance" to group and it will be refactorable

src/main/java/school/hei/haapi/endpoint/rest/controller/GroupController.java
```diff
@@ -30,12 +31,10 @@ public Group getGroupById(@PathVariable String id) {
   }
 
   @GetMapping(value = "/groups")
-  public List<Group> getGroups(
+  public GroupStat getGroups(
```
Mahefaa: nope, it doesn't return the GroupStatistics, it returns the Group, find a better name

src/main/java/school/hei/haapi/endpoint/rest/mapper/UserMapper.java
```diff
@@ -32,6 +37,16 @@ public UserIdentifier toIdentifier(User user) {
         .email(user.getEmail());
   }
 
+  public StudentStat toRestStudentStat(List<User> students) {
```
Mahefaa: StudentStat is not what you should return

Mahefaa (changes requested):

PR #535, opened 2024-07-17
feat: filter students by exclude group id

Iloniavo:


doc/api.yml
```diff
@@ -458,6 +458,12 @@ paths:
           schema:
             type: string
             format: date-time
+        - name: exclude_group_id
```
Mahefaa: rename to excluded_groups and use a string array

Mahefaa (changes requested):
where are the tests ?

src/test/java/school/hei/haapi/integration/StudentIT.java
```diff
@@ -666,6 +681,18 @@ void manager_write_update_ok() throws ApiException {
     assertTrue(updated.contains(updated1));
   }
 
+  @Test
+  void manager_read_student_by_exclude_group_id() throws ApiException {
+    ApiClient manager1Client = anApiClient(MANAGER1_TOKEN);
+    UsersApi api = new UsersApi(manager1Client);
+
+    List<Student> students =
+        api.getStudents(1, 10, null, null, null, null, null, null, null, null, "group1_id");
```
Mahefaa: add one blank line after the WHEN

Mahefaa (changes requested):

doc/api.yml
```diff
@@ -458,6 +458,14 @@ paths:
           schema:
             type: string
             format: date-time
+        - name: exclude_groups
+          required: false
+          in: query
+          description: List of group ids in which students are not in
```
Mahefaa: list of group ids to exclude students from

Mahefaa (changes requested):
last change before merge

doc/api.yml
Mahefaa: what happens if the exclude_groups is empty ?

PR #561, opened 2024-08-07
feat(not-implemented): add field for repeating student

Anjaraniaina:


doc/api.yml
```diff
@@ -3796,6 +3796,8 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Group'
+        is_repeating:
+          type: boolean
```
haritianaadriano: is_repeating_year is more explicit, i think

haritianaadriano (changes requested):

PR #562, opened 2024-08-07
feat(not-implemented): add endpoint for courses with teacher

mirado447:


doc/api.yml
```diff
@@ -653,7 +653,76 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
-
+  /courses_with_teacher:
+    get:
+      tags:
+        - Teaching
+      summary: Get all courses with teacher
+      operationId: getCoursesWithTeacher
```
haritianaadriano: I think get all courses assign to a teacher is more consistant

haritianaadriano: same here

haritianaadriano (changes requested):

doc/api.yml
Mahefaa, line 1: rejected because you're forcing Teachers to tell us the entirety of the courses they teach right at their doorstep.

A possible enhancement could be to do that on a specific endpoint and cover special cases like two teachers teaching the same course but during different periods in time or even teaching the same course at the same period

Mahefaa (changes requested):

doc/api.yml
```diff
@@ -1224,7 +1305,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Teacher'
+                $ref: '#/components/schemas/CrupdateTeacher'
```
Iloniavo: Why CrupdateTeacher instead of Teacher model ?

doc/api.yml
```diff
@@ -1318,7 +1399,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: '#/components/schemas/Teacher'
+                  $ref: '#/components/schemas/CrupdateTeacher'
```
Iloniavo: same here

doc/api.yml
```diff
@@ -3799,9 +3880,6 @@ components:
     Teacher:
       allOf:
         - $ref: '#/components/schemas/CrupdateUser'
-      properties:
-        profile_picture:
-          type: string
```
Iloniavo: why is this removed ?

Iloniavo (changes requested):

doc/api.yml
mirado447: my fault, I correct taht

PR #563, opened 2024-08-08
feat: filter for security conf to disable user suspended

haritianaadriano:


src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
```diff
@@ -80,24 +80,26 @@ req, res, null, forbiddenWithRemoteInfo(req))))
...
             AnonymousAuthenticationFilter.class)
 
         // authorize
         .authorizeHttpRequests(
             request ->
                 request
-                    .requestMatchers(GET, "/ping")
+                    .requestMatchers(
+                        new AntPathRequestMatcher("/ping", GET.name()),
+                        new AntPathRequestMatcher("/ping", GET.name()))
```
Mahefaa: two ping endpoints ?

requestMatchers(GET, "/ping") was already correct

Mahefaa: you need to write down all ENDPOINTS which require authentication, implying those with .hasAnyRole or .hasRole because before getting someone's role, you have to know their identity (hence authentication)

Mahefaa (changes requested):

src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
```diff
@@ -55,6 +55,76 @@ public AuthenticationManager authenticationManager() {
     return new ProviderManager(authProvider);
   }
 
+  RequestMatcher MANAGER_AUTHORIZATION_PATH =
```
Mahefaa: rename to MANAGER_ONLY_REQUEST_MATCHER

as static variable if possible and private!!!


do the same for the others.

only selfMatchers can't be static as they need authResourceProvider field

src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
```diff
@@ -80,24 +150,28 @@ req, res, null, forbiddenWithRemoteInfo(req))))
...
+                    antMatcher(GET, "/students/*/files"),
+                    antMatcher(GET, "/students/*/scholarship_certificate/raw"))),
             AnonymousAuthenticationFilter.class)
 
         // authorize
         .authorizeHttpRequests(
             request ->
                 request
-                    .requestMatchers(GET, "/ping")
+                    .requestMatchers(new AntPathRequestMatcher("/ping", GET.name()))
```
Mahefaa: no need to use the AntPathRequestMatcher here

src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
Mahefaa, line 1: for self-matchers, you will have to duplicate the matchers as you must not do the self-matching on bearerFilter  because they are not yet authenticated.

For that purpose, create a constructor which takes a RequestMatcher and an AuthResourceProvider for each SelfMatchers

Mahefaa (changes requested):

PR #565, opened 2024-08-09
feat: implementation of repeating student of current year

Anjaraniaina:


src/main/java/school/hei/haapi/model/Promotion.java
```diff
@@ -31,10 +32,11 @@ public class Promotion {
   private String id;
 
   @CreationTimestamp private Instant creationDatetime;
-
   private String ref;
   private String name;
 
+  private LocalDate startDate;
+
   @OneToMany(mappedBy = "promotion")
```
haritianaadriano: Use Instant lika other model class

src/main/java/school/hei/haapi/model/User.java
```diff
@@ -105,6 +105,8 @@ public class User implements Serializable {
 
   private String highSchoolOrigin;
 
+  private transient Boolean isRepeatingYear;
+
```
haritianaadriano: i think you can use the primitve type not the Boolean type Class

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -233,4 +242,38 @@ public Integer getStudentsAlternatingSize(
...
+  }
+
+  public Boolean didRepeatYear(User user) {
+    int normalNumberOfPromotion = 1;
+    Set<Promotion> distinctPromotions = getStudentDistinctPromotions(user);
+    return distinctPromotions.size() > normalNumberOfPromotion;
+  }
+
+  public Set<school.hei.haapi.model.Promotion> getStudentDistinctPromotions(User user) {
+    List<school.hei.haapi.model.Group> belongedGroup =
```
haritianaadriano: Your method is not clear, i think you can rename this to getStudentCurrentPromotion or getAllStudentPromotion

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -233,4 +242,38 @@ public Integer getStudentsAlternatingSize(
...
+  public Boolean computeIsRepeatingThisYear(User user) {
+    Set<school.hei.haapi.model.Promotion> distinctPromotions = getStudentDistinctPromotions(user);
+    if (distinctPromotions.isEmpty()) {
+      return false;
+    }
+    school.hei.haapi.model.Promotion mostRecentPromotion =
+        distinctPromotions.stream()
+            .sorted(Comparator.comparingInt(prom -> prom.getStartDate().getYear()))
+            .toList()
+            .getLast();
```
haritianaadriano: i think you can sort by creation datetime of the promotion flow or group flow by DESC

haritianaadriano: reduce heavy treatment with java if it's possible, and did it during sql treatment for more fluidity and performance

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -233,4 +242,38 @@ public Integer getStudentsAlternatingSize(
...
+    return didRepeatYear(user)
+        && (mostRecentPromotion.getStartDate().getYear() == Year.now().getValue()
+            || mostRecentPromotion.getStartDate().getYear() + 1 == Year.now().getValue());
```
haritianaadriano: why you add + 1 when you calculate it ?

haritianaadriano (changes requested):

src/main/java/school/hei/haapi/service/UserService.java
Anjaraniaina: in test mock data, all creation datetime are the same for all promotion so I assumed that it's not the start date of the promotion

src/main/java/school/hei/haapi/service/UserService.java
Anjaraniaina: Suppose current year is 2023 and the most recent promotion start date year is 2023, how do I know if a student repeat a year ? I confirm that he has at least 2 prom, then if the most recent Promotion's start date Year == this year (2023) then the student repeated a year, 
but what if the current year is 2024 ?
Promotion's start date Year + 1 == this year (2023+1 == 2024) then the student repeated a year because a year in college is like this 2023-2024
Does it touch the promotion 2024-2025 ?
if current year is 2023, then it's not but if current year is 2024, then I just need to add a condition that the prom start year + 1 (2025 here) <= current year

PR #585, opened 2024-08-27
infra: add new user role as MONITOR

haritianaadriano:


src/test/java/school/hei/haapi/integration/SecurityIT.java
```diff
@@ -49,6 +50,14 @@ public static Whoami whoisStudent1() {
     return whoami;
   }
 
+  public static Whoami whoisMonitor1() {
+    Whoami whoami = new Whoami();
+    whoami.setId("monitor1_id");
+    whoami.setBearer(MONITOR1_TOKEN);
+    whoami.setRole(Whoami.RoleEnum.MONITOR);
```
Mahefaa: static import

Mahefaa (approved):
looks good to me except a little static import.
However having the same table for monitors, users, students, managers now seems a bit weird.
We could store Student specific attributes to another table and so on for all of the roles

PR #588, opened 2024-08-28
feat(not-implemented): letterbox resources

Iloniavo:


doc/api.yml
```diff
@@ -3275,6 +3275,163 @@ paths:
...
+          application/json:
+            schema:
+              type: array
+              items:
+                type: string
+                description: List of deposits ids
+              example: ["deposit_1_id", "deposit_2_id"]
+      responses:
+        '200':
+          description: All deposit in the global letterbox
```
Mahefaa: description is wrong

doc/api.yml
```diff
@@ -3275,6 +3275,163 @@ paths:
...
+          $ref: '#/components/responses/404'
+        '429':
+          $ref: '#/components/responses/429'
+        '500':
+          $ref: '#/components/responses/500'
+    put:
+      tags:
+        - Letterbox
+      summary: Update deposits status
+      operationId: updateDespositsStatus
```
Mahefaa: update depositStatus should have a status in its requestbody, otherwise it simply means that you've read the deposit and the operation would be readDeposit

doc/api.yml
```diff
@@ -3275,6 +3275,163 @@ paths:
         '500':
           $ref: '#/components/responses/500'
 
+  '/letterbox':
```
Mahefaa: path naming and return type naming is non-consistent

doc/api.yml
```diff
@@ -3275,6 +3275,163 @@ paths:
...
+          required: false
+        - name: student_ref
+          in: query
+          description: Filter by student ref
+          schema:
+            type: string
+          required: false
+      responses:
+        '200':
+          description: All deposit in the global letterbox
```
Mahefaa: it's not all of them, it's a paginated list

doc/api.yml
```diff
@@ -3275,6 +3275,163 @@ paths:
...
+          description: Set value to 15 by default if null is provided
+          schema:
+            $ref: '#/components/schemas/PageSize'
+          required: false
+        - name: student_ref
+          in: query
+          description: Filter by student ref
+          schema:
+            type: string
+          required: false
```
Mahefaa, line 3302: create components for page and pageSize parameters, you're allowed to do this in another PR but note it

doc/api.yml
```diff
@@ -3275,6 +3275,163 @@ paths:
...
+      responses:
+        '200':
+          description: All deposit in the global letterbox
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Deposit'
```
Mahefaa: returning array as top-level response is not good for possible changes.

Could you be the first one to use some sort of PageResponse ?

PagedResponse:
data : type array
pageSize: number
page: number
hasPrevious : boolean => page >1

Mahefaa (changes requested):

PR #593, opened 2024-08-30
feat(not-implemented): monitor resources

haritianaadriano:


doc/api.yml
```diff
@@ -2973,7 +3155,7 @@ paths:
             $ref: '#/components/schemas/PageSize'
         - name: from
           in: query
-          description: Filter events from a given date, return events of current week by default
+          description: Filter events from a given date, return events of current week by defaultfalse
```
Mahefaa: should be unchanged

doc/api.yml
Mahefaa, line 1: how to unlink students ?

Mahefaa (changes requested):

doc/api.yml
haritianaadriano: i think we will see later

PR #604, opened 2024-09-05
chore: check exam resources

mirado447:


haritianaadriano:
Are you sure that this functionality is ok

haritianaadriano (commented):
I renamed your PR because it's not reffer to what you did actually, and it's a chore because you check the exam resources.

PR #668, opened 2024-10-22
feat(not-implemented): get monthy fees statistics

mirado447:


doc/api.yml
```diff
@@ -2007,6 +2007,16 @@ paths:
               - UNPAID
               - LATE
               - PAID
+        - name: start_month
```
haritianaadriano: month_from is better, from and to are both temporaly marks

doc/api.yml
```diff
@@ -2007,6 +2007,16 @@ paths:
...
+          in: query
+          schema:
+            type: string
+            format: date
+        - name: end_month
```
haritianaadriano: month_to (same here)

haritianaadriano: but the type is always instant so use the date-time format, this is the convention in the application

doc/api.yml
```diff
@@ -2044,7 +2054,30 @@ paths:
...
+      tags:
+        - Paying
+      summary: Get monthly statistics on Fees
+      parameters:
+        - name: start_month
+          in: query
+          schema:
+            type: string
+            format: date
+        - name: end_month
```
haritianaadriano: same here use the month_from and month_to, add "the actual month by default" in your description. We will return the stats of the month by default if both of them are not provided

doc/api.yml
haritianaadriano: here we don't need to specify the month, we will just return all by default with pagination

haritianaadriano (changes requested):

PR #755, opened 2024-12-12
chore: not suspend pending student

tendryAxel:


src/main/java/school/hei/haapi/service/MpbsService.java
```diff
@@ -33,4 +34,8 @@ public Mpbs getByPspId(String pspId) {
         .findByPspId(pspId)
         .orElseThrow(() -> new NotFoundException("Mpbs not found #" + pspId));
   }
+
+  public Long countPendingOfStudent(String studentId) {
+    return mpbsRepository.countMpbsByStatusAndStudentId(MpbsStatus.PENDING, studentId);
```
Iloniavo, line 39: use static import for enum

Iloniavo (changes requested):

src/main/java/school/hei/haapi/service/MpbsService.java
tendryAxel: I can't, there is already a static import of FeeStatusEnum.PENDING in the class

PR #758, opened 2024-12-17
feat: GET /staff_members/xlsx

kirisaki-vk:


doc/api.yml
```diff
@@ -2015,6 +2015,33 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+
+  /staff_members/xslx/raw:
```
haritianaadriano: It's a good practice to specify the returned type before the /raw as you do but by convention in hei admin use directly /staff_member/raw

doc/api.yml
```diff
@@ -2015,6 +2015,33 @@ paths:
...
+    get:
+      tags:
+        - Users
+      summary: Get all staff members as xlsx file
+      operationId: getStaffMembersXlsx
```
haritianaadriano: Say explicitly, the method must be named by what it did: getStaffMembersXlsx is too vague try something getStaffMembersIntoXlsx

src/main/java/school/hei/haapi/endpoint/rest/controller/StaffMemberController.java
```diff
@@ -50,6 +57,12 @@ public List<StaffMember> getStaffMembers(
         .toList();
   }
 
+  @GetMapping("/staff_members/xlsx/raw")
+  public byte[] getStaffMembersXlsx() {
+    return userService.getByRoleAndStatusAsXlsx(
+        STAFF_MEMBER, User.Status.ENABLED, userMapper::toRestStaffMember);
```
haritianaadriano: Static import

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -153,6 +154,28 @@ public List<User> getByRoleAndStatus(User.Role role, User.Status status) {
     return userRepository.findAllByRoleAndStatus(role, status);
   }
 
+  public <T> byte[] getByRoleAndStatusAsXlsx(
+      User.Role role, User.Status status, Function<User, T> mapper) {
+    List<User> users = getByRoleAndStatus(role, status);
+    XlsxCellsGenerator<T> generator = new XlsxCellsGenerator<>();
+    List<T> mappedUsers = users.stream().map(mapper).toList();
```
haritianaadriano, line 167: Where is your test ? I think you can directly use the STAFF_MEMBER User type here, also don't forget to test your code before merging. 

How we will work:
.1 write the api.doc (indeed)
.2 if all are ok we will merge the spec and generate the client. Then Frontend can do their work part
.3 and it's after all that we will implement

haritianaadriano: + Test

haritianaadriano (changes requested):

doc/api.yml
kirisaki-vk: Changing to `/staff_member/raw` will match with the `/staff_member/{id}` endpoint 
Should I change it to `/staff_member/raw/xlsx` then ?

src/main/java/school/hei/haapi/service/UserService.java
kirisaki-vk: Understood
I'm still working on the tests
For some reason the endpoint `/staff_member/xlsx/raw` always returns me forbiden even by adding the proper rules in the `SecurityConf`

doc/api.yml
kirisaki-vk: Ok, I will change it to `getStaffMemberIntoXlsx`

doc/api.yml
kirisaki-vk: I changed it to `/staff_member/raw/xlsx`

src/main/java/school/hei/haapi/service/UserService.java
kirisaki-vk: Tests are now done

haritianaadriano (approved):

PR #764, opened 2024-12-18
chore: GET /students multiple status filter

kirisaki-vk:


doc/api.yml
```diff
@@ -463,7 +463,9 @@ paths:
           in: query
           description: Filter students by status value return all by default
           schema:
-            $ref: '#/components/schemas/EnableStatus'
+            type: array
+            items:
+              $ref: '#/components/schemas/EnableStatus'
```
haritianaadriano, line 468: I think it's time to rename this variable. And use a better name like UserStatus or StudentStatus, EnableStatus means only ENABLE as status no ?

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
```diff
@@ -77,13 +77,15 @@ public List<Student> getStudents(
...
+      @RequestParam(name = "status", required = false, defaultValue = "ENABLED,SUSPENDED")
+          List<EnableStatus> status,
       @RequestParam(name = "sex", required = false) Sex sex,
       @RequestParam(name = "work_study_status", required = false) WorkStudyStatus workStatus,
       @RequestParam(name = "commitment_begin_date", required = false) Instant commitmentBeginDate,
       @RequestParam(name = "exclude_groups", required = false) List<String> excludeGroupIds) {
     User.Sex domainSex = sexEnumMapper.toDomainSexEnum(sex);
-    User.Status domainStatus = statusEnumMapper.toDomainStatus(status);
+    List<User.Status> domainStatus =
+        status == null ? null : status.stream().map(statusEnumMapper::toDomainStatus).toList();
```
haritianaadriano, line 88: handle it in the mapper method, and why you give a default value as ENABLED, SUSPENDED but you handle nullable value here ?

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -214,9 +214,10 @@ public List<User> getByCriteria(
       User.Sex sex) {
     Pageable pageable =
         PageRequest.of(page.getValue() - 1, pageSize.getValue(), Sort.by(ASC, "ref"));
+    List<User.Status> statusList = status == null ? null : List.of(status);
```
haritianaadriano, line 217: why you handle always the null value. But ok, i'll be fine with it but remember sometimes it's a bad things to use null as value instead it's better to use Optional or List and check if it's empty if a List and check if present if Optional

src/test/java/school/hei/haapi/integration/StudentIT.java
```diff
@@ -1016,11 +1016,50 @@ void stats_are_exact() throws ApiException {
...
-        usersApi.getStudents(1, 200, null, null, null, null, null, null, null, null, null).size();
+        usersApi
+            .getStudents(
+                1,
+                200,
+                null,
+                null,
+                null,
+                null,
+                List.of(SUSPENDED, ENABLED, DISABLED),
```
haritianaadriano, line 1057: and where you check Student with only two status ?

haritianaadriano (changes requested):

src/main/java/school/hei/haapi/service/UserService.java
Mahefaa: if status null, List.of(), else List.of(status) a List object should never be null

doc/api.yml
Mahefaa: agreed, however it will be a breaking change

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
Mahefaa: agreed, also list variables should be named with plural nouns, hence statuses
2) a list should never be null but empty or populated.

src/main/java/school/hei/haapi/repository/dao/UserManagerDao.java
```diff
@@ -29,7 +29,7 @@ public List<User> findByCriteria(
       String firstName,
       String lastName,
       Pageable pageable,
-      User.Status status,
+      List<User.Status> status,
```
Mahefaa, line 32: userStatuses, static import User.Status

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
```diff
@@ -77,13 +77,15 @@ public List<Student> getStudents(
       @RequestParam(value = "first_name", required = false, defaultValue = "") String firstName,
       @RequestParam(value = "last_name", required = false, defaultValue = "") String lastName,
       @RequestParam(value = "course_id", required = false, defaultValue = "") String courseId,
-      @RequestParam(name = "status", required = false) EnableStatus status,
+      @RequestParam(name = "status", required = false, defaultValue = "ENABLED,SUSPENDED")
+          List<EnableStatus> status,
```
Mahefaa, line 81: userStatuses

src/main/java/school/hei/haapi/service/UserService.java
```diff
@@ -227,7 +228,7 @@ public List<User> getByLinkedCourse(
       String courseId,
       PageFromOne page,
       BoundedPageSize pageSize,
-      User.Status status,
+      List<User.Status> status,
```
Mahefaa, line 231: userStatuses, static import User.Status

src/main/java/school/hei/haapi/endpoint/rest/controller/StudentController.java
Mahefaa: param name can be the same for retro-compatibility however you MUST change your variable names

Mahefaa (changes requested):

PR #782, opened 2025-01-16
feat: deny scholarship certificate on missing nic number

kirisaki-vk:


src/main/java/school/hei/haapi/service/StudentFileService.java
```diff
@@ -112,8 +113,15 @@ public FileInfo getUserFileById(String userId, String id) {
     return fileInfoRepository.getByUserIdAndId(userId, id);
   }
 
-  public byte[] generatePdf(String studentId, String template) {
+  public byte[] generateScholarshipCertificate(String studentId, String template) {
     Context context = loadContext(studentId);
+    User student = userService.findById(studentId);
+    int studentAge = now().getYear() - student.getBirthDate().getYear();
+    if (studentAge >= 18 && student.getNic() == null) {
```
haritianaadriano: Use instead a validator and userIdentifier or Principal class to take the connected user

haritianaadriano (changes requested):

src/main/java/school/hei/haapi/service/StudentFileService.java
kirisaki-vk: Solved it in the the latest commit

src/main/java/school/hei/haapi/service/StudentFileService.java
```diff
@@ -112,8 +114,9 @@ public FileInfo getUserFileById(String userId, String id) {
     return fileInfoRepository.getByUserIdAndId(userId, id);
   }
 
-  public byte[] generatePdf(String studentId, String template) {
+  public byte[] generateScholarshipCertificate(String studentId, String template) {
     Context context = loadContext(studentId);
+    ageValidator.accept(18);
```
haritianaadriano: Why you use a static valor as 18 instead of taking the age of the concerned user ?

haritianaadriano: And by convention we use the validator in the controller

haritianaadriano (changes requested):

src/main/java/school/hei/haapi/service/StudentFileService.java
kirisaki-vk: If I take the age of the user I have no use for the Principal in the validator

I have refactored it and made so it takes the user from the studenId path parameter and also make the validator to take the user in need of validation not the Principal.

haritianaadriano (approved):

PR #787, opened 2025-01-17
feat: mail on participant missing event

kirisaki-vk:


src/main/resources/templates/missedEventEmail.html
```diff
@@ -0,0 +1,156 @@
...
+            . À défaut, vous serez sujet à des sanctions
+            pouvant aller jusqu'à une Convocation, Observation ou Renvoi (COR).
+
+        </p>
+    </section>
+    <section>
+        Pour toute question ou besoin d’assistance, n’hésitez pas à contacter
+        l’administration à l’adresse
+        <strong>contact@hei.school</strong> ou au numéro
+        <strong>+261 34 94 041 16</strong>.
```
haritianaadriano, line 132: Ask our client if this body pass with what they expected

kirisaki-vk: I asked them what the body of the mail should look like and they gave me that template

src/main/java/school/hei/haapi/service/EventParticipantService.java
```diff
@@ -43,11 +47,20 @@ public List<EventParticipant> getEventParticipants(
   }
 
   public List<EventParticipant> updateEventParticipants(List<EventParticipant> eventParticipants) {
+    List<EventParticipant> missingParticipants =
+        eventParticipants.stream()
+            .filter(
+                participant -> {
+                  return participant.getStatus().equals(MISSING);
```
haritianaadriano: Better to begin with the value that you are sure to exists. For example it's better to assert something like: `MISSING.equals(participant.getStatus());`

haritianaadriano: to handle null value and prevent throwing

haritianaadriano (changes requested):

src/main/java/school/hei/haapi/service/EventParticipantService.java
kirisaki-vk: Oh ! That's true

haritianaadriano: Why it's allays from the student status

haritianaadriano (changes requested):

src/main/java/school/hei/haapi/service/EventParticipantService.java
kirisaki-vk: This has been moved to the appropriate Service and it has already been changed in the latest commit

PR #814, opened 2025-01-30
feat(not-implemented): get event statistics

kirisaki-vk:


doc/api.yml
```diff
@@ -4245,6 +4245,54 @@ paths:
...
+          description: the eventId of the event if you want to see the stats for a specific event
+          required: false
+          schema:
+            type: string
+        - name: from
+          required: false
+          in: query
+          description: |
+            specify the date of start of the event. 
+            If `to` is specified but this parameter is left blank, the default value will be the current date_time
```
haritianaadriano: your description is not clear, try something like that: from event datetime to fetch, by default the current date_time. Descibre the meaning of the parameter and what the parameter does, not what client must do

doc/api.yml
```diff
@@ -4245,6 +4245,54 @@ paths:
...
+          schema:
+            type: string
+            format: date-time
+        - name: to
+          required: false
+          description: |
+            specify the date of start of the event. 
+            If `from` is specified but this parameter is left blank, the default value will be the current date_time
```
haritianaadriano: same

doc/api.yml
```diff
@@ -4245,6 +4245,54 @@ paths:
...
+          in: query
+          schema:
+            type: string
+            format: date-time
+      responses:
+        '200':
+          description: Stats info
```
haritianaadriano: stats of what ?

haritianaadriano (changes requested):

doc/api.yml
Mahefaa, line 1: close unused PR

PR #831, opened 2025-02-13
fix: correct count of event create by frequency

tendryAxel:


src/main/java/school/hei/haapi/service/utils/DatetimeProducer.java
```diff
@@ -18,10 +18,9 @@ public class DatetimeProducer implements BiFunction<List<LocalDate>, String, Lis
   public List<Instant> apply(List<LocalDate> localDates, String time) {
     DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern("HH:mm");
     LocalTime localTime = LocalTime.parse(time, timeFormatter);
-    ZoneId zoneId = ZoneId.of("UTC+3");
```
Mahefaa, line 21: why did you change it ?

src/test/java/school/hei/haapi/integration/EventIT.java
```diff
@@ -111,10 +114,45 @@ void manager_create_event_and_event_participant_by_frequence_ok() throws ApiExce
...
+    assertEquals(Instant.parse("2023-12-27T12:00:00Z"), eventWeek3.getEndDatetime());
+  }
+
+  @Test
+  void manager_modify_ok() throws ApiException {
+    ApiClient apiClient = anApiClient(MANAGER1_TOKEN);
+    EventsApi api = new EventsApi(apiClient);
+
+    CreateEvent eventCourse1 = createEventCourse1();
+    List<Event> actual = api.crupdateEvents(List.of(eventCourse1), null, null, null, null);
```
Mahefaa: add blank line
Given
WHEN
THEN

Mahefaa (changes requested):
where is the bug ? what does the fix do ? 
write a tests which fails when the bug happens and show that it passes

src/main/java/school/hei/haapi/service/utils/DatetimeProducer.java
tendryAxel: it uses the wrong zone id, when converted to Instant it is no longer correct
You can see before that if it uses UTC+3 here, there's a mismatch between the excepted and the actual.

Mahefaa (approved):
approved, but please ensure that using UTC as timezone is the correct way, you guys seem to use UTC+3 quite often

PR #834, opened 2025-02-19
fix: handle null comment on fee advanced stats

kirisaki-vk:


src/main/java/school/hei/haapi/model/Fee.java
```diff
@@ -166,20 +166,28 @@ public int hashCode() {
   }
 
   public boolean isWorkStudyStudentFee() {
-    return this.getComment().toLowerCase().contains("alternance") && TUITION.equals(this.getType());
+    return Optional.ofNullable(this.getComment())
+            .orElse("Unknown")
+            .toLowerCase()
+            .contains("alternance")
+        && TUITION.equals(this.getType());
```
Mahefaa: TUITION.equals first


Optional.ofNullable(str).map(str contains alternance).orElse(false)

src/main/java/school/hei/haapi/model/Fee.java
```diff
@@ -166,20 +166,28 @@ public int hashCode() {
...
+            return L1;
+          }
+          if (comment.toLowerCase().contains("l2")) {
+            return L2;
+          }
+          if (comment.toLowerCase().contains("l3")) {
+            return L3;
+          }
+          return null;
+        });
```
Mahefaa: If null return empty() else return switch on comment, 3 cases contains L1 or L2 or L3. default empty()

Mahefaa (changes requested):

src/main/java/school/hei/haapi/model/Fee.java
kirisaki-vk: If the mapping function returns a null result then this method returns an empty Optional

src/main/java/school/hei/haapi/model/Fee.java
```diff
@@ -166,20 +169,28 @@ public int hashCode() {
   }
 
   public boolean isWorkStudyStudentFee() {
-    return this.getComment().toLowerCase().contains("alternance") && TUITION.equals(this.getType());
+    return TUITION.equals(this.getType())
+        && Optional.ofNullable(this.getComment())
+            .map(comment -> comment.toLowerCase().contains("alternance"))
```
Mahefaa: alternance keyword

src/main/java/school/hei/haapi/model/Fee.java
```diff
@@ -191,12 +202,16 @@ public PaymentType getPaymentType() {
...
-      return Optional.of(MONTHLY);
-    }
-    if (this.getComment().toLowerCase().contains("annuel")) {
-      return Optional.of(YEARLY);
-    }
-    return Optional.empty();
+    Optional<String> optionalComment = Optional.ofNullable(this.getComment());
+    return optionalComment.map(
+        comment -> {
+          if (comment.toLowerCase().contains(MONTHLY_FEE_KEYWORD)) {
```
Mahefaa: comment.toLowerCase() in a var

Mahefaa (changes requested):

Mahefaa (approved):

PR #837, opened 2025-02-21
chore: creation event by frequency

tendryAxel:


src/main/java/school/hei/haapi/service/EventService.java
```diff
@@ -140,6 +140,7 @@ private List<Event> generateEventFromFrequency(
   private List<Event> duplicateEventWithDifferentHour(
       Event event, Map<Instant, Instant> eventScheduled) {
     List<Event> duplicated = new ArrayList<>();
+    duplicated.add(event);
```
Mahefaa, line 143: this line was removed on our last review, why does it need to get back ?

Mahefaa (changes requested):

doc/api.yml
```diff
@@ -4236,7 +4236,10 @@ paths:
       tags:
         - Events
       summary: Create or update events
-      description: Update if id is provided, create otherwise.
+      description: |
+        Update if id is provided, create otherwise.
+        
+        **IMPORTANT** The request must be send as UTC+3
```
Mahefaa: Not using timezone is the best way to avoid timezone issues.

You use Instant as your datetime, and consumers use it too. If they want to show the instant at a specific timezone, they apply the timezone changes onto the instant

Mahefaa (changes requested):

src/main/java/school/hei/haapi/service/EventService.java
tendryAxel: I misunderstand the usage of the functionality, it needs to create one supplementary event for the frequency

Mahefaa (approved):

PR #841, opened 2025-02-26
feat(not-implemented): announcements reaction

tendryAxel:


doc/api.yml
```diff
@@ -5169,6 +5169,35 @@ paths:
...
+      tags:
+        - Announcements
+      summary: toggle reaction for an announcement
+      parameters:
+        - name: id
+          required: true
+          schema:
+            type: string
+          in: path
+      operationId: reactToAnnouncement
```
Mahefaa, line 5183: PUT operations must have request body

Mahefaa (changes requested):

doc/api.yml
```diff
@@ -6587,6 +6626,13 @@ components:
           format: date-time
         scope:
           $ref: '#/components/schemas/Scope'
+        reaction_count:
+          type: number
+    ReactionEnum:
+      type: string
+      enum:
+        - UNCHECK
+        - CHECK
```
Mahefaa, line 6635: Reactions will be persisted for us to compute their count, to ensure idempotency, add an id to the Reaction

Mahefaa (changes requested):

doc/api.yml
tendryAxel: But the reaction depend on the announcement id and user id.
To be idempotent, all you need is the id of the announce and the user

PR #851, opened 2025-03-13
chore: moved sequence generation as preprocessing event

kirisaki-vk:


src/main/java/school/hei/haapi/endpoint/event/model/HandleReceiptGenerationRequest.java
```diff
@@ -0,0 +1,35 @@
...
+
+@AllArgsConstructor
+@EqualsAndHashCode
+@ToString
+@Data
+@Builder
+public class HandleReceiptGenerationRequest extends PojaEvent {
+
+  @JsonProperty("payments")
+  private List<Payment> payments;
```
Mahefaa, line 21: there should be a limit on how many payments can be handled by one event, add a TODO and you'll work on it if the problem ever appears

src/main/java/school/hei/haapi/service/PaymentNumberSequenceService.java
```diff
@@ -13,6 +13,7 @@ public class PaymentNumberSequenceService {
   private final PaymentNumberSequenceRepository repository;
 
   public PaymentNumberSequence getNextSequence(LocalDate date) {
+    // TODO: Make getNextSequence able to handle concurrent calls.
```
Mahefaa, line 16: is the method still used?

src/main/java/school/hei/haapi/service/event/SendRequestReceiptGenerationService.java
```diff
@@ -15,13 +14,20 @@
...
+              sendRequestReceiptGeneration.getPayment());
+      receiptGenerationService.saveReceipt(
+          feeReceiptGenerated, sendRequestReceiptGeneration.getPayment());
+      log.info("Payment: {} are generated", sendRequestReceiptGeneration.getPayment());
+    } catch (Exception e) {
+      log.error(
+          "Receipt generation failed for {} with error : {}",
+          sendRequestReceiptGeneration.getPayment(),
+          e.getMessage());
+    }
```
Mahefaa: ok, i get what you want to do, but there are a few Exceptions which should be handled differently,

Failed events are meant to be retried, if the cause of failure is something that can be fixed through retries, like s3 failed upload due to network error

src/test/java/school/hei/haapi/unit/sequence/PaymentSequenceTest.java
Mahefaa, line 1: nice

Mahefaa (changes requested):

src/main/java/school/hei/haapi/service/PaymentNumberSequenceService.java
kirisaki-vk: This method is still use by PaymentService to get the next available sequence if the payment passed doesn't have a sequence yet

Mahefaa: it's fine then

Mahefaa (approved):

PR #862, opened 2025-03-24
chore: add category and frequency to fees

kirisaki-vk:


src/main/resources/db/migration/V44_67__Add_fee_category_and_frequency.sql
```diff
@@ -0,0 +1,18 @@
+do
+$$
+    begin
+        if not exists(select from pg_type where typname = 'fee_category') then
+            create type "fee_category" as enum ('L1','L2','L3','REMEDIAL_COSTS', 'WORK_FEES', 'UNKNOWN', 'EMPTY_GF');
```
Mahefaa: EMPTY_GF?

Mahefaa (approved):

src/main/resources/db/migration/V44_67__Add_fee_category_and_frequency.sql
kirisaki-vk: This tells the category of the fee cannot be determined because the User who made it has no Group Flows to get the category

PR #879, opened 2025-04-15
fix: getAllEventParticipants without date parameters

tendryAxel:


src/main/java/school/hei/haapi/repository/dao/EventParticipantDao.java
```diff
@@ -25,6 +25,7 @@
 @AllArgsConstructor
 public class EventParticipantDao {
   private final EntityManager entityManager;
+  private static final String EVENT_BEGIN_DATETIME = "beginDatetime";
```
Mahefaa: could be public static in Event class

Mahefaa (changes requested):
i do not see any fix in the PR

tendryAxel:
The issue is EventController::getAllEventParticipants don't work without "from" and "to" filter

src/main/java/school/hei/haapi/repository/dao/EventParticipantDao.java
tendryAxel: Can't because Event is a JPA Entity

Mahefaa: public static allows exports from other packages, also,  BEGIN_DATETIME is an attribute of class Event.
Have a quick look at open api generated code.
image

Mahefaa: name it JSON_PROPERTY_BEGIN_DATETIME or anything along those lines

src/main/java/school/hei/haapi/repository/dao/EventParticipantDao.java
```diff
@@ -108,16 +109,17 @@ private List<Predicate> getPredicates(
       predicates.add(builder.equal(root.get("status"), attendanceStatus));
     }
 
+    predicates.add(builder.isNotNull(root.get("event").get(EVENT_BEGIN_DATETIME)));
```
Mahefaa: root.get("event") in a variable and re-use it, or event root.get("event").get(blabla);

Mahefaa (changes requested):

Mahefaa (approved):

PR #884, opened 2025-05-08
fix: student status going from disabled to suspended

kirisaki-vk:


src/main/java/school/hei/haapi/model/User.java
```diff
@@ -151,6 +150,11 @@ public void setStatus(Status newStatus) {
     if (DISABLED.equals(this.getStatus()) && ENABLED.equals(newStatus)) {
       throw new IllegalArgumentException("DISABLED User cannot be set back to ENABLED");
     }
+
+    if (DISABLED.equals(this.getStatus()) && SUSPENDED.equals(newStatus)) {
+      throw new IllegalArgumentException("DISABLED User cannot be set to SUSPENDED");
+    }
```
Mahefaa, line 156: can we even do anything to disabled student ??

src/test/java/school/hei/haapi/unit/SuspendStudentWithOverdueFeesTest.java
```diff
@@ -0,0 +1,51 @@
...
+    var subject =
+        new SuspendStudentsWithOverdueFeesService(userManagerDao, userService, mpbsService);
+
+    when(userService.getStudentsWithUnpaidOrLateFee()).thenReturn(disabledStudents);
+    when(mpbsService.countPendingOfStudent(any())).thenReturn(0L);
+
+    subject.suspendStudentsWithUnpaidOrLateFee();
+    var student = userRepository.findById(DISABLED_STUDENT_ID).orElse(null);
+
+    assertEquals(User.Status.DISABLED, student.getStatus());
```
Mahefaa: static import

Mahefaa (changes requested):

src/main/java/school/hei/haapi/model/User.java
kirisaki-vk: We got a case where a scheduler fetched a students even if they are disabled 
We've fixed it but to avoid cases like these in the future we found it better to put that check

Mahefaa (approved):

PR #888, opened 2025-05-14
fix: advanced fee stats null stat_date on update

kirisaki-vk:


Mahefaa (changes requested):
do not request review when tests are not passing

kirisaki-vk:
> do not request review when tests are not passing

It should pass but I'm figuring it out for now 
I'll request again when all is ok

src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
```diff
@@ -94,6 +89,11 @@ private Collection<AdvancedFeeStats> generateAdvancedFeeStatsOnDate(LocalDate da
           });
       return stats;
     } else {
+      dayStat.replaceAll(
+          (statType, stat) -> {
+            stat.setId(UUID.randomUUID().toString());
```
Mahefaa: hmm, why don't your stats have an ID before this operation?

kirisaki-vk: These are newly generated stats for the day so they have no Id yet and also we can't mark the id field as Generated since it caused problems while updating the values for existing ones

Mahefaa: hmm still dubious, why wouldn't you newly generated states have an ID that you give (at creation time, with the builder if NOT-GENERATED, and at creation time via uuid_generate_v4 if GENERATED) ?

also for your algorithm's optimization, there's no need to generateAdvancedFeeStats  before feeDao.getAdvancedFeeStatsOnDate.isEmpty()

Mahefaa (changes requested):

Mahefaa (approved):

PR #895, opened 2025-05-21
refactor: replace verbose code in manager_write_with_some_bad_fields_ko

tokyramarozaka:
refactored manager_write_with_some_bad_fields_ko test which was too long and had many redundancies. I simply used a helperFunction `assertApiException` which does both `assertThrows` and `assertEquals` on the expected error message.


Mahefaa (approved):
ok, but we already have a method named AssertThrowsApiException somewhere in TestUtils

Mahefaa (changes requested):
Code needs to be formatted with our formatter, checkout format.sh

tokyramarozaka:
okay, good, will do. I saw the format CI failing yes. I didn't think that we already had an `assertApiException` because I thought the previous guys would have been smart enough to use it already.

tokyramarozaka:
@Mahefaa fixed. I have more tests to refactor, but this was just to check how things are done here.

Mahefaa (approved):

PR #901, opened 2025-05-30
chore: implement exams grade stats for a student in a course

tendryAxel:


src/main/java/school/hei/haapi/service/ExamService.java
```diff
@@ -93,4 +94,15 @@ public List<Exam> getAllExams(
...
     return examRepository.findExamsByCourseId(courseId);
   }
+
+  public StudentCourseGradeStats getStudentCourseGradeStats(String courseId, String studentId) {
+    List<Exam> exams = getExamsByCourseId(courseId);
+    double sumScore =
+        gradeService.scoreSumWithCoefficient(
+            exams.stream()
+                .map(exam -> gradeService.getGradeByExamIdAndStudentId(exam.getId(), studentId))
+                .toList());
```
Mahefaa: too many sql calls, you could create a method with a custom query to get exam and grade within a single sql, and only call getGradedExamsByCourseId

Mahefaa (changes requested):

PR #902, opened 2025-05-30
fix: advanced fee stats get late and pending filter

kirisaki-vk:


kirisaki-vk:
This seems to be already fixed

PR #903, opened 2025-05-30
feat: filter student grades by academic year level

tendryAxel:


src/main/java/school/hei/haapi/model/Group.java
```diff
@@ -86,4 +92,19 @@ public boolean equals(Object o) {
   public int hashCode() {
     return getClass().hashCode();
   }
+
+  public StudentGrade getStudentGradeAtInstant(LocalDate localDate) {
+    int promotionYearAge =
+        Period.between(LocalDate.from(this.getPromotion().getStartDatetime()), localDate)
+            .getYears();
```
Mahefaa, line 99: i think we've had some bugs concerning this, like not knowing the exact promotion at a certain time

src/test/java/school/hei/haapi/integration/GradeIT.java
Mahefaa, line 1: you're not testing your new filter at all

Mahefaa (changes requested):

PR #910, opened 2025-06-11
test: missing participant

tendryAxel:


src/main/java/school/hei/haapi/model/Event.java
```diff
@@ -57,42 +60,30 @@ public class Event {
   // TODO : Add promotion
 
   @Column(name = "begin_datetime")
+  @ToStringExclude
```
Mahefaa: @ToString.Exclude exists, why would you use another library's

src/test/java/school/hei/haapi/endpoint/event/CheckParticipantMissingEventTest.java
```diff
@@ -0,0 +1,81 @@
...
+            eq(null), any(Instant.class), any(Instant.class), eq(null), eq(null), eq(null)))
+        .thenReturn(List.of(event1));
+    when(eventParticipantService.findByEventId(anyString(), eq(null)))
+        .thenReturn(
+            List.of(
+                new EventParticipant("participant1", event1, new User(), MISSING, null),
+                new EventParticipant("participant2", event1, new User(), MISSING, null)));
+
+    // This function modify nothing, just test if not throw
+    assertDoesNotThrow(() -> subject.accept(new CheckParticipantMissedEventTriggered()));
```
Mahefaa: ok it does not modify anything, but you could test if it does what you want it to do, other than not throwing, like does it send an event, does it read something from db, and test the final objective, not the steps in between

Mahefaa (changes requested):

Mahefaa (approved):

PR #916, opened 2025-06-24
chore: add advanced fee stats invalidation

kirisaki-vk:


src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
```diff
@@ -76,7 +82,7 @@ public AdvancedFeesStatistics getAdvancedFeeStats(
     LocalDate from = Optional.ofNullable(dateFrom).orElse(now.withDayOfMonth(1));
     LocalDate to = Optional.ofNullable(dateTo).orElse(now.with(lastDayOfMonth()));
     var advancedStats = feeDao.getAdvancedFeeStatsOnDateBetween(from, to, type.orElse(RECEIPT));
-    if (advancedStats.isEmpty()) {
```
Mahefaa, line 79: hmmm, ok if shouldBeUpdated then we go for an update but the said update should be done in async,
if the required update takes too much time because there is too much data, the task will timeout and update will never succeed.

What you could do is, send the async event and then, try to compute the result, if result computing somehow fails, (takes more than 10 seconds let's say), you indicate that async computation already started and sync computation couldn't do it on time

Mahefaa: some sort of fallible eager loading

Mahefaa: also please correct line 83, use of LocalDatetime.MAX sometimes fails in db, as db.MAX might not equal Java's, use atTime(23,59,59) instead

Mahefaa: BUT, do not forget to update "updateDatetime" field, in order to avoid running multiple async updates on the same resource because previous async updates didn't run yet

Mahefaa: or some sort of status haha

Mahefaa (changes requested):

src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
kirisaki-vk: I've done it in async now

doc/api.yml
```diff
@@ -7146,6 +7146,12 @@ components:
           $ref: "#/components/schemas/PendingFeesStats"
         late_fees_count:
           $ref: "#/components/schemas/LateFeesStats"
+        update_datetime:
+          type: string
+          format: date-time
+        expired:
+          type: boolean
+          description: Ths field will be `true` when the returned stats is outdated or needs to be generated.
```
Mahefaa, line 7154: needs to be generated ?

src/main/java/school/hei/haapi/service/event/AdvancedFeeStatsComputationTriggeredService.java
```diff
@@ -25,7 +25,7 @@ public void accept(AdvancedFeeStatsComputationTriggered advancedFeeStatsComputat
     Optional<Instant> fromValue =
         Optional.ofNullable(advancedFeeStatsComputationTriggered.getBeginDatetime().toInstant(UTC));
     Optional<Instant> toValue =
-        Optional.ofNullable(advancedFeeStatsComputationTriggered.getEndDatetime().toInstant(UTC));
```
Mahefaa, line 28: i don't get why you rename your attributes all the time

Mahefaa (approved):

doc/api.yml
kirisaki-vk: Yeah, If there is no stats for the selected date yet then it needs to be generated at first

kirisaki-vk: generation will be done by the first request

PR #924, opened 2025-07-23
docs: added yearly_results and course_grades endpoints

kirisaki-vk:


doc/api.yml
```diff
@@ -477,6 +477,69 @@ paths:
...
+  /students/{student_id}/yearly_results:
+    get:
+      tags:
+        - Grades
+      summary: Get the yearly results of a given student for all levels
+      operationId: getYearlyResults
+      parameters:
+        - $ref: "#/components/parameters/PathStudentId"
+        - name: student_level
+          required: false
```
Mahefaa: why isn't it required?

kirisaki-vk: Not specifying it will return the result for all of the year

doc/api.yml
```diff
@@ -477,6 +477,69 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /students/{student_id}/yearly_results:
+    get:
+      tags:
+        - Grades
+      summary: Get the yearly results of a given student for all levels
```
Mahefaa: yearly results? plural noun returns single result

doc/api.yml
Mahefaa: do we really need to return everything instead of forcing a student level filter ?

tokyramarozaka: it is situational, but for the end of the year we need to calculate all three years (like we did for you guys). we initially thought about forcing a student level filter here, and having another endpoint for all three years, perhaps that will differentiate it more. As the three years is one-time feature at the end.

doc/api.yml
tokyramarozaka: this ties with the other change : if we only want to return a single result, then yes, totally will always return a single yearly result.

doc/api.yml
tokyramarozaka: @kirisaki-vk maybe create another endpoint like /summary for all three years and enforce a student level filter on this one ?

tokyramarozaka:
Just wondering.... is the results_summary ok now ? @Mahefaa @kirisaki-vk

doc/api.yml
```diff
@@ -6446,6 +6544,37 @@ components:
           type: number
           format: double
           example: 12.5
+    CourseResult:
+      type: object
+      properties:
+        id:
+          type: string
+        couse:
```
Mahefaa: typo: course

doc/api.yml
```diff
@@ -477,6 +477,94 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /students/{student_id}/yearly_results/{student_level}:
+    get:
+      tags:
+        - Grades
+      summary: Get the yearly results of a given student for the given level.
+      operationId: getYearlyResults
```
Mahefaa: getYearlyResult

doc/api.yml
```diff
@@ -477,6 +477,94 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /students/{student_id}/yearly_results/{student_level}:
+    get:
+      tags:
+        - Grades
+      summary: Get the yearly results of a given student for the given level.
```
Mahefaa: result

Mahefaa (changes requested):

Mahefaa:
Courses need to be linked to the StudentLevel, it's the only way to filter them by student levels anyway

doc/api.yml
kirisaki-vk: Yep It's settled now

Mahefaa (approved):

PR #926, opened 2025-07-26
feat: add getCourseGrades to fetch grades for a student for a given course

tokyramarozaka:
This feat returns a list of grades obtained by a student at each exam, for a single course.


tokyramarozaka:
Upon doing this PR, I have set many many todos, that I think are worth sharing : 
- we need a permanent way to deal with creation timestamps: @CreationTimestamp overwrites the value you set on tests sometimes. Maybe try PrePersist to say : if the creationTimestamp is null, then set it to now(), else, leave it there. 
- we should add GradeHistory next.

src/main/java/school/hei/haapi/endpoint/rest/controller/GradeController.java
```diff
@@ -8,61 +8,45 @@
...
 
   @PutMapping(value = "/exams/{exam_id}/students/{student_id}/grade")
-  public school.hei.haapi.endpoint.rest.model.Grade crupdateParticipantGrade(
+  public Grade crupdateParticipantGrade(
       @PathVariable("exam_id") String examId,
       @PathVariable("student_id") String studentId,
       @RequestBody CrupdateGrade grade) {
     validator.accept(grade);
-    Grade toSave = gradeMapper.toDomain(grade, examId, userService.findById(studentId).getRef());
+    school.hei.haapi.model.Grade toSave =
```
Mahefaa: use var

src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
```diff
@@ -60,7 +56,9 @@ public AuthenticationManager authenticationManager() {
   }
 
   @Bean
-  public SecurityFilterChain configure(HttpSecurity httpSecurity) throws Exception {
+  public SecurityFilterChain configure(
+      HttpSecurity httpSecurity, MonitoringStudentService monitoringStudentService)
```
Mahefaa: why isn't MonitoringStudentService autowired like the other injected services ?

src/main/java/school/hei/haapi/service/GradeService.java
```diff
@@ -24,7 +26,14 @@
 public class GradeService {
   private final GradeRepository gradeRepository;
   private final GradeDao gradeDao;
-  private final UserService userService;
+  private final UserRepository userRepository;
+  private final CourseAssignmentRepository courseAssignmentRepository;
+
+  public List<Grade> getGradesByStudentId(String studentId) {
+    var student = userRepository.getById(studentId);
+    log.info("getGradesByStudentId: {}", student);
```
Mahefaa: rm or use log.debug

Mahefaa (changes requested):
Great, will be approved once you resolve these issues

src/main/java/school/hei/haapi/service/GradeService.java
tokyramarozaka: removed.

tokyramarozaka:
for the record: it is done. @Mahefaa

Mahefaa (approved):

PR #931, opened 2025-08-03
feat: add getStudentByIdAndMonitorId feature with tests

tokyramarozaka:
We needed an endpoint for monitors to check a single student that they monitor, so here we are. I added tests for the service in case : 
a. the student does not exist
b. the monitor does not actually have such student.

### DIFF
--- a/src/main/java/school/hei/haapi/service/MonitoringStudentService.java
+++ b/src/main/java/school/hei/haapi/service/MonitoringStudentService.java
@@ -1,6 +1,5 @@
 package school.hei.haapi.service;
 
-import static java.util.stream.Collectors.toUnmodifiableList;
 import static org.springframework.data.domain.Sort.Direction.ASC;
 
 import java.util.List;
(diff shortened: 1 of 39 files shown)


tokyramarozaka:
@kirisaki-vk I constantly have the same error as you. 

* What went wrong:
Execution failed for task ':compileJava'.
> java.lang.StackOverflowError

any ideas why ?

src/main/java/school/hei/haapi/service/MonitoringStudentService.java
```diff
@@ -110,4 +110,17 @@ private User mapMonitorToDomain(CrupdateMonitor monitor) {
...
+
+  public User getStudentByIdAndMonitorId(String studentId, String monitorId)
+      throws BadRequestException {
+    var optionalStudent = userRepository.findById(studentId);
+    if (optionalStudent.isEmpty() || optionalStudent.get().getRole() != User.Role.STUDENT) {
+      throw new BadRequestException("Student with id: " + studentId + " does not exist");
+    }
+    if (!monitoringStudentRepository.getAllMonitorsIdsByStudentId(studentId).contains(monitorId)) {
+      throw new BadRequestException(
+          "Monitor with id: " + monitorId + " does not have a student with id: " + studentId);
```
Mahefaa, line 122: throw NotFoundException, it is more suitable.

Mahefaa (changes requested):

PR #936, opened 2025-08-08
feat: YearlyResult transcript

kirisaki-vk:


doc/api.yml
```diff
@@ -507,6 +507,36 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /students/{student_id}/yearly_results/{student_level}/raw:
+    get:
+      tags:
+        - Grades
+      summary: Get the yearly result of a given student for the given level into a pdf file.
+      operationId: getYearlyResultFile
```
Mahefaa: operationId lets us think that this returns a File when it returns a Status

doc/api.yml
```diff
@@ -6613,6 +6643,18 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/CourseResult"
+    YearlyResultGenerationState:
+      type: object
+      properties:
+        status:
+          $ref: "#/components/schemas/YearlyResultGenerationStatus"
+        link:
+          type: string
```
Mahefaa: component is never used

doc/api.yml
```diff
@@ -507,6 +507,36 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /students/{student_id}/yearly_results/{student_level}/raw:
```
Mahefaa: /raw is used for a raw file, like a byte[] or a base64, not for json

Mahefaa (changes requested):

doc/api.yml
```diff
@@ -507,6 +507,36 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /students/{student_id}/yearly_results/{student_level}/transcript:
+    get:
+      tags:
+        - Grades
+      summary: Get the yearly result generation state of a given student for the given level.
+      operationId: getYearlyResultTranscriptState
```
Mahefaa: getYearlyResultTranscript, you get the transcript with its state not a state

Mahefaa (changes requested):

doc/api.yml
kirisaki-vk: Oh 
You're correct

Mahefaa (changes requested):
bad summary and response 200 description

Mahefaa (approved):

PR #941, opened 2025-08-13
chore: add tests and advancedFeeStats query for receipt type

tokyramarozaka:
Generates advanced fee stats for the RECEIPT type. The other one being the ACCOUNTING type which was implemented. 

For further reference on the difference between the two, please refer to the following conversation. 

image

### DIFF
--- a/src/main/java/school/hei/haapi/repository/FeeRepository.java
+++ b/src/main/java/school/hei/haapi/repository/FeeRepository.java
@@ -70,4 +70,6 @@ public interface FeeRepository extends JpaRepository<Fee, String> {
       @Param(value = "toCompare") Instant toCompare,
       @Param("studentId") String studentId,
       @Param("status") FeeStatusEnum status);
+
+  List<Fee> findAllByStatusHistoriesDatetimeBetween(Instant dayStart, Instant dayEnd);
 }
--- a/src/main/java/school/hei/haapi/repository/FeeRepository.java
+++ b/src/main/java/school/hei/haapi/repository/FeeRepository.java
@@ -71,5 +71,5 @@ public interface FeeRepository extends JpaRepository<Fee, String> {
       @Param("studentId") String studentId,
       @Param("status") FeeStatusEnum status);
 
-  List<Fee> findAllByStatusHistoriesDatetimeBetween(Instant dayStart, Instant dayEnd);
+  List<Fee> findDistinctByStatusHistoriesDatetimeBetween(Instant dayStart, Instant dayEnd);
 }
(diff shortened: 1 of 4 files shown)


src/main/java/school/hei/haapi/repository/FeeRepository.java
```diff
@@ -70,4 +70,6 @@ List<Fee> getStudentFeesUnpaidOrLateFrom(
       @Param(value = "toCompare") Instant toCompare,
       @Param("studentId") String studentId,
       @Param("status") FeeStatusEnum status);
+
+  List<Fee> findAllByStatusHistoriesDatetimeBetween(Instant dayStart, Instant dayEnd);
```
Mahefaa: so, we're assuming that 01 jan 00:00:00.00Z-> 02 Jan 00:00:00Z is a correct filter?

src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
```diff
@@ -127,7 +127,11 @@ private Collection<AdvancedFeeStats> generateAdvancedFeeStatsOnDateBetween(
     Instant dayStart = fromDate.atStartOfDay().toInstant(UTC);
     Instant dayEnd = toDate.atTime(23, 59, 59).toInstant(UTC);
 
-    List<Fee> allFees = feeRepository.findAllByDueDatetimeBetween(dayStart, dayEnd);
+    List<Fee> allFees =
+        switch (type) {
+          case ACCOUNTING -> feeRepository.findAllByDueDatetimeBetween(dayStart, dayEnd);
+          case RECEIPT -> feeRepository.findAllByStatusHistoriesDatetimeBetween(dayStart, dayEnd);
```
Mahefaa: is it possible for a fee to have multiple statuses on a date-range ?

e.g:
 PENDING, and PAID on the same month for example, is your code able to handle such scenario?

src/main/java/school/hei/haapi/repository/FeeRepository.java
Mahefaa: or must it be a full month date-range?

tokyramarozaka:
I should probably find a better name for this PR, reading it again, it looks horrible.

src/main/java/school/hei/haapi/repository/FeeRepository.java
tokyramarozaka: ` 01 jan 00:00:00.00Z-> 02 Jan 00:00:00Z is a correct filter?`
-> yes it is, the stats for that date range shall be generated, but if you try to access an already existing date-range, then it won't be generated again, it will simply be fetched.

src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
tokyramarozaka: yes, we fetch the LAST status, the closest one to the TO instant. perhaps, a test for that is needed ?

Mahefaa: yes, we do need one

src/main/java/school/hei/haapi/repository/FeeRepository.java
Mahefaa: ok

src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
tokyramarozaka: @Mahefaa there's already a test for that it in FeeStatusAtTest.java actually. We already use getStatusAt(Instant t), and it is already tested.

Mahefaa (approved):

PR #964, opened 2025-08-28
fix: handle runtime exceptions in mpbs verifications

kirisaki-vk:
Mpbs verification is done in asynchronous and when a verification of one mpbs fails, we should continue for others not stop the entire process

### DIFF
--- a/src/main/java/school/hei/haapi/service/MpbsVerificationService.java
+++ b/src/main/java/school/hei/haapi/service/MpbsVerificationService.java
@@ -80,43 +80,44 @@ public class MpbsVerificationService {
                   Comparator.comparing(
                       MobileTransactionDetails::getPspDatetimeTransactionCreation));
 
-      if (correspondingTransactionPendingDetails.isPresent()) {
-        MobileTransactionDetails lastTransactionDetails =
-            correspondingTransactionPendingDetails.get();
-        if (SUCCESS.equals(lastTransactionDetails.getStatus())) {
-          try {
-            log.info("mobile transaction found = {}", lastTransactionDetails);
-            TransactionDetails transactionDetails =
-                externalResponseMapper.toExternalTransactionDetails(lastTransactionDetails);
-            log.info("mapped transaction details = {}", transactionDetails);
-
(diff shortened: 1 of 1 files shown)


src/main/java/school/hei/haapi/service/MpbsVerificationService.java
```diff
@@ -97,6 +97,11 @@ public List<MpbsVerification> verifyMobilePaymentAndSaveResult(List<Mpbs> pendin
                 "payment %s could not be verified because fee %s has no remaining amount"
                     .formatted(pendingMbps.getId(), pendingMbps.getFee().getId()),
                 e);
+          } catch (RuntimeException e) {
+            log.error(
+                "Mpbs of ref {} could not be verified because of error {}",
+                pendingMbps.getPspId(),
+                e.getMessage());
```
Mahefaa: bad way of logging exception, 
https://medium.com/w-logs/how-to-log-exception-properly-6aa80b62ff8a
it's useless if Exception is of type NullPointerException because NullPointerException's message is "null"

Mahefaa (changes requested):

src/main/java/school/hei/haapi/service/MpbsVerificationService.java
```diff
@@ -80,40 +80,44 @@ public List<MpbsVerification> verifyMobilePaymentAndSaveResult(List<Mpbs> pendin
...
+      if (!SUCCESS.equals(lastTransactionDetails.getStatus())) {
+        log.info(
+            "verification mobile payment details stored is not success for the payment {}",
+            pendingMbps.getId());
+        unverifiedMpbs.add(pendingMbps);
+        continue;
+      }
+
+      try {
+        log.info("mobile transaction found = {}", lastTransactionDetails);
```
Mahefaa: is it a log.debug?

kirisaki-vk: Probably 
It was always there

Mahefaa (approved):

PR #987, opened 2025-09-12
fix(remedial): restructure remedial component

ghost:


kirisaki-vk (approved):
Wait for tests and merge if it passes

PR #1002, opened 2025-09-23
feat: add retake exam inscription date

ghost:


doc/api.yml
```diff
@@ -7360,6 +7358,9 @@ components:
           $ref: "#/components/schemas/Course"
         session:
           $ref: "#/components/schemas/RetakeExamSession"
+        inscription_date:
```
hei-ryan: `registration_date` instead of `inscription_date`

doc/api.yml
```diff
@@ -7360,6 +7358,9 @@ components:
...
+          type: string
+          format: datetime
```
hei-ryan: I think the correct format is `date-time` not `datetime`

hei-ryan (changes requested):

doc/api.yml
ghost: yes, I'm changing it

doc/api.yml
ghost: Already fixed

PR #1003, opened 2025-09-23
feat: add retake exam registration date and refactor test

ghost:


doc/api.yml
```diff
@@ -7360,6 +7358,9 @@ components:
           $ref: "#/components/schemas/Course"
         session:
           $ref: "#/components/schemas/RetakeExamSession"
+        inscription_date:
```
hei-ryan: Same remarks as in another PR : registration instead of inscription.

src/main/java/school/hei/haapi/model/RetakeExam.java
```diff
@@ -44,4 +45,6 @@ public class RetakeExam implements Serializable {
   @ManyToOne
   @JoinColumn(name = "student_id", nullable = false)
   private User student;
+
+  private Instant inscriptionDate;
```
hei-ryan: By default, set current_timestamp.

src/main/java/school/hei/haapi/service/RetakeExamService.java
```diff
@@ -58,10 +59,14 @@ private Boolean isExisted(CrupdateRetakeExam crupdateRetakeExam) {
   public List<RetakeExam> getStudentRetakeExams(String sessionId, String studentId) {
     var session = retakeExamSessionService.getById(sessionId);
     var coursesToRetake = getCourseResultToRetake(studentId, session);
-
-    return coursesToRetake.stream()
-        .map(courseResult -> courseResultAndSessionToRetake(courseResult, session))
-        .toList();
+    var existedRetakeExams = getExistedRetakeExams(session, studentId);
```
hei-ryan: existed is not a appropriate word, use existing instead. 

Always correct your translation using deepl.com

src/main/java/school/hei/haapi/service/RetakeExamService.java
```diff
@@ -76,6 +81,11 @@ private RetakeExam courseResultAndSessionToRetake(
     return retakeExam;
   }
 
+  private List<RetakeExam> getExistedRetakeExams(RetakeExamSession session, String studentId) {
+    return retakeExamRepository.findRetakeExamsBySession_IdAndStudent_Id(
```
hei-ryan: Why did you create a specific function for this as it is called only once and it's just a fetch from database ? Juste use the retakeExamRepository.findRetakeExamsBySession_IdAndStudent_Id method where it's needed.

src/main/resources/db/migration/V45_98__add_retake_exam_inscription_date_column.sql
```diff
@@ -0,0 +1,2 @@
+alter table "retake_exam"
+    add column if not exists inscription_date timestamp;
```
hei-ryan: Do not forget to rename it.

src/test/resources/db/testdata/V99_35__Test_create_retake_exam.sql
```diff
@@ -1,4 +1,4 @@
-insert into "retake_exam" (id, session_id, course_id, student_id)
-values('retake_exam1_id', 'session1_id', 'course1_id', 'student1_id'),
-      ('retake_exam2_id', 'session1_id', 'course2_id', 'student2_id'),
-      ('retake_exam3_id', 'session1_id', 'course3_id', 'student3_id');
\ No newline at end of file
+insert into "retake_exam" (id, session_id, course_id, student_id, inscription_date)
+values('retake_exam1_id', 'session1_id', 'course1_id', 'student1_id','2025-09-11T08:00:00.00Z' ),
```
hei-ryan: This way of having test is bad, even if it is actually the way tests were written before. What you should do is to create your data on your own test class, here for example in RetakeExamIT and delete created data after each test, so it does not influence other tests that may use RetakeExam entities. Just take note of it, but for now as you need to do it as quick as possible, leave it as it.

src/main/java/school/hei/haapi/endpoint/rest/mapper/RetakeExamMapper.java
```diff
@@ -33,21 +35,25 @@ public school.hei.haapi.model.RetakeExam toDomainCrupDate(CrupdateRetakeExam cru
...
         .build();
   }
 
   public RetakeExam toRest(school.hei.haapi.model.RetakeExam retakeExam) {
     return new RetakeExam()
         .id(retakeExam.getId())
         .course(courseMapper.toRest(retakeExam.getCourse()))
-        .session(retakeExamSessionMapper.toRest(retakeExam.getSession()));
+        .session(retakeExamSessionMapper.toRest(retakeExam.getSession()))
+        .inscriptionDate(String.valueOf(retakeExam.getInscriptionDate()));
```
hei-ryan: No. You use String here only because you map badly the format in the API spec. Instead of format: date-time, you wrote format: datetime. 

Actually, the appropriate type here, once you'll do the change, will directly be Instant type.

hei-ryan (changes requested):

src/main/java/school/hei/haapi/model/RetakeExam.java
ghost: Already, changed

src/main/resources/db/migration/V45_98__add_retake_exam_inscription_date_column.sql
ghost: Yes, I changed it directly in the same file because it's not in preprod yet.

doc/api.yml
ghost: yes, I already changed it

src/main/java/school/hei/haapi/service/RetakeExamService.java
ghost: Yes, it's noted

src/main/java/school/hei/haapi/service/RetakeExamService.java
ghost: I think using specific functions is easier to read. Because if we put all these functions inside one function, it will become disorganized.

src/test/resources/db/testdata/V99_35__Test_create_retake_exam.sql
ghost: Yes, It's noted

src/main/java/school/hei/haapi/endpoint/rest/mapper/RetakeExamMapper.java
ghost: I found it

src/test/java/school/hei/haapi/integration/RetakeExamSessionIT.java
```diff
@@ -42,8 +41,8 @@ void setUp() {
   }
 
   @Test
-  void read_all_retake_exam_by_student_ok() throws ApiException {
-    ApiClient apiClient = anApiClient(STUDENT1_TOKEN);
+  void read_all_retake_exam_sessions_by_admin_ok() throws ApiException {
+    ApiClient apiClient = anApiClient(ADMIN1_TOKEN);
     RetakeExamApi api = new RetakeExamApi(apiClient);
 
     var retakeAxamSessions = api.getRetakeExamSessions(null, null, null, null, null);
```
kirisaki-vk, line 60: Don't rename the tests 
Leave the test for student_ok and make another one for admin to checks for premissions

kirisaki-vk (changes requested):

src/test/java/school/hei/haapi/integration/RetakeExamSessionIT.java
ghost: yes, I'm going to create another test in that case

src/main/java/school/hei/haapi/endpoint/rest/mapper/RetakeExamMapper.java
```diff
@@ -33,21 +34,24 @@ public school.hei.haapi.model.RetakeExam toDomainCrupDate(CrupdateRetakeExam cru
...
+        .session(retakeExamSessionMapper.toRest(retakeExam.getSession()))
+        .registrationDate(retakeExam.getRegistrationDate());
   }
 
   public school.hei.haapi.model.RetakeExam toDomain(RetakeExam retakeExam) {
     return school.hei.haapi.model.RetakeExam.builder()
         .id(retakeExam.getId())
         .course(courseMapper.toDomain(retakeExam.getCourse()))
         .session(retakeExamSessionMapper.toDomain(retakeExam.getSession()))
+        .registrationDate(Instant.now())
```
hei-ryan: Can't you make it only through database ? 

So its value will be really null before persistence and current timestamp value after the persistence ? In that way, we really can consider that it was generated only after persistence. So here, just really map retakeExam.getRegistrationDate

hei-ryan (changes requested):

src/main/java/school/hei/haapi/endpoint/rest/mapper/RetakeExamMapper.java
ghost: Yes, I can do it through the database. 

So, I'll remove all default value with instant.now()

kirisaki-vk (approved):

PR #1004, opened 2025-09-24
feat(doc): add retake exam inscription date

ghost:


tendryAxel (approved):

doc/api.yml
```diff
@@ -1514,14 +1514,12 @@ paths:
           required: true
           schema:
             type: string
-        - $ref: '#/components/parameters/TeacherId'
```
hei-ryan, line 1517: Why did you remove it here ?

doc/api.yml
```diff
@@ -1514,14 +1514,12 @@ paths:
...
         - name: course_code
           in: query
           required: false
           example: PROG1
           schema:
             type: string
-        - $ref: "#/components/parameters/GroupRef"
```
hei-ryan, line 1524: Why did you remove it here ?

I mean through this PR

hei-ryan (changes requested):

doc/api.yml
ghost: we use Course instead of CourseAssignment. Since Course doesn’t have a groupRef, it’s not possible to filter by groupRef. In this case, the student is treated as a UserIdentifier, and UserIdentifier also doesn’t contain a groupRef.

doc/api.yml
ghost: Because we use Course instead of CourseAssignment. Since Course doesn’t have a teacher, it’s not possible to filter by teacherId.

hei-ryan (approved):
Do not forget to rename the commit from inscription to registraiton

PR #1014, opened 2025-09-25
chore: add verification date before creating a session

ghost:


src/main/java/school/hei/haapi/service/RetakeExamSessionService.java
```diff
@@ -33,6 +33,9 @@ public List<RetakeExamSession> getRetakeExamSessions(
   }
 
   public RetakeExamSession save(RetakeExamSession retakeExamSession) {
+    if (!retakeExamSession.getDateFrom().isBefore(retakeExamSession.getDateTo())) {
```
tendryAxel: `!....isBefore()` is the same as `....isAfter()`

ghost: yes, I'll use it

tendryAxel (approved):

PR #1015, opened 2025-09-26
test: crupdate retake exam session with wrong date

ghost:


tendryAxel (commented):
There is a good way to make exceptions in a backend.
It would be easier to show you directly how it's done.

src/main/java/school/hei/haapi/service/RetakeExamSessionService.java
```diff
@@ -33,9 +34,8 @@ public List<RetakeExamSession> getRetakeExamSessions(
   }
 
   public RetakeExamSession save(RetakeExamSession retakeExamSession) {
-    if (retakeExamSession.getDateFrom().isAfter(retakeExamSession.getDateTo())) {
-      throw new IllegalArgumentException("Session start date must be before end date");
-    }
+    if (retakeExamSession.getDateFrom().isAfter(retakeExamSession.getDateTo()))
+      throw new BadRequestException("Session start date must be before end date");
```
tendryAxel, line 38: as you can see here, we have an Exception called `BadRequestException`, instead of returning a 500 error, it'll be a 400 error, because the mistake is from the user not the server

src/test/java/school/hei/haapi/integration/RetakeExamSessionIT.java
```diff
@@ -75,4 +77,17 @@ void save_retake_exam_session_by_admin() throws ApiException {
...
+  void save_retake_exam_session_with_wrong_date_ko() {
+    ApiClient apiClient = anApiClient(ADMIN1_TOKEN);
+    RetakeExamApi api = new RetakeExamApi(apiClient);
+    var sessionWithWrongDate = sessionWithWrongDate();
+
+    assertBadRequestException(
+        "Session start date must be before end date",
+        () ->
+            api.createOrUpdateRetakeExamSessions(
+                retakeExamSessionMapper.toRest(sessionWithWrongDate)));
```
tendryAxel, line 91: `TestUtils::assertBadRequestException` is a simple and custom assertion function to test if the api create a bad request exception

src/main/java/school/hei/haapi/service/RetakeExamSessionService.java
ghost: Yes, got it

src/test/java/school/hei/haapi/integration/RetakeExamSessionIT.java
ghost: got it,

tendryAxel (approved):

PR #1027, opened 2025-10-03
fix: verify all pending mpbs on xlsx verification

kirisaki-vk:


kirisaki-vk:
This PR is being closed since the change has already been made directly in prod

PR #1028, opened 2025-10-03
feat(doc): add retake exam status

ghost:


doc/api.yml
```diff
@@ -1520,6 +1520,12 @@ paths:
           example: PROG1
           schema:
             type: string
+        - name: status
+          in: query
+          required: false
+          example: CANCELED
+          schema:
+            type: string
```
tendryAxel: Make that a list for more flexibility
Like by default, the front can make the admin see REGISTERED && TO_CANCEL Retake at the same time

tendryAxel: like that
```
    CorStatuses:
      name: cor_status
      in: query
      required: false
      description: Filter by cor actual status
      schema:
        type: array
        items:
          $ref: '#/components/schemas/CorStatus'
```
But i'm not sure if that work

tendryAxel (changes requested):

doc/api.yml
ghost: Wait, that's not a COR, that's a retake exam or should I just make a list filter like this?

ghost: changed

doc/api.yml
```diff
@@ -6749,6 +6750,15 @@ components:
...
         type: string
+    RetakeExamStatuses:
+      name: retake_exam_status
+      in: query
+      required: false
+      description: Filter by retake exam actual status
+      schema:
+        type: array
+        items:
+          $ref: '#/components/schemas/RetakeExamStatus'
```
tendryAxel, line 6761: why not

tendryAxel (approved):

PR #1033, opened 2025-10-07
refactor(doc):  rename retakeExams to retake_exams and add new endpoint

ghost:


doc/api.yml
```diff
@@ -1634,6 +1704,11 @@ paths:
           required: true
           schema:
             type: string
+        - name: std_ref
+          in: query
+          required: false
+          schema:
+            type: string
```
tendryAxel: replace with `- $ref: '#/components/parameters/StudentRef'`

doc/api.yml
```diff
@@ -1618,7 +1652,43 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
-  /retake_exam_sessions/{session_id}/retake_exam_courses/{course_id}/students:
+  /retake_exam_sessions/{session_id}/retake_exam_courses/{course_id}:
+    get:
+      tags:
+        - RetakeExam
+      summary: Get retake exam course by course identifier
```
tendryAxel: `and session identifier`

tendryAxel: But i don't understand why do you need the `session_id` if you already have the `course_id`.
Maybe you don't need the `/retake_exam_sessions`

doc/api.yml
```diff
@@ -1618,7 +1652,43 @@ paths:
...
+      operationId: getRetakeExamCourseById
+      responses:
+        '200':
+          description: Retake exam course by Id
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Course'
```
tendryAxel: why `getRetakeExamCourseById` and then return `Course`, you must return `RetakeExam`, it's the more natural response here

tendryAxel (changes requested):

doc/api.yml
ghost: Okay, get it

doc/api.yml
ghost: Session_id is a just a path
And retake_exam_sessions is just for all session existed

doc/api.yml
ghost: here the client expect to return the a specific course with because he need it when he show the information of this course.

ghost: We need only the course inside retake exam but not the retake exam

doc/api.yml
tendryAxel: Wait do you mean `course id` ? If so you need to change the endpoint name like `.../courses/{course_id}`
But if you mean `rekate_exam(or course I don't really remember)`, you need to delete the `/retake_exam_session` because it's not necessary for this case

tendryAxel (changes requested):
The purpose of the endpoint isn't clear

doc/api.yml
ghost: Ah, yeah, I understand, so why not use the endpoint getCourseById because, this endpoint return too the course

tendryAxel (approved):

PR #1036, opened 2025-10-08
feat(not-implemented): cor interviewers

tendryAxel:


doc/api.yml
```diff
@@ -8394,13 +8394,21 @@ components:
       properties:
         concerned_student_id:
           type: string
+        interviewers_id:
```
kirisaki-vk: interviewer_ids ?

kirisaki-vk (changes requested):

PR #1070, opened 2025-11-04
feat(doc): add student level for session and filter retake exam by student ref

DyferHerioss:


doc/api.yml
```diff
@@ -7500,6 +7504,10 @@ components:
           $ref: "#/components/schemas/Course"
         session:
           $ref: "#/components/schemas/RetakeExamSession"
+        student_comment:
+          type: string
+        validator_comment:
+          type: string
```
kirisaki-vk, line 7510: These are outside of the scope of this PR 
Put this in a new PR, also I'm not satisfied by this name. It would be discussed there

doc/api.yml
```diff
@@ -7518,6 +7526,10 @@ components:
           type: string
         status:
           $ref: "#/components/schemas/RetakeExamStatus"
+        student_comment:
+          type: string
+        validator_comment:
+          type: string
```
kirisaki-vk, line 7532: These values should be updated in a separate endpoint

kirisaki-vk (changes requested):

PR #1073, opened 2025-11-05
feat: (doc) add retake exam comments

DyferHerioss:


doc/api.yml
```diff
@@ -7503,6 +7503,10 @@ components:
         registration_date:
           type: string
           format: date-time
+        student_comment:
+          type: string
+        validator_comment:
+          type: string
```
kirisaki-vk: Again these should have a different name
Student comment and Validator comment doesn't seem right if they should represent the cancel request and the request rejection cause

doc/api.yml
```diff
@@ -7516,6 +7520,10 @@ components:
           type: string
         student_id:
           type: string
+        student_comment:
+          type: string
+        validator_comment:
+          type: string
```
kirisaki-vk: These values should be updated through a different endpoint not by crupdating the whole RetakeExam

kirisaki-vk (changes requested):

doc/api.yml
DyferHerioss: But it not only change retake exam cancelation reason, it can change too the retake exam state

doc/api.yml
```diff
@@ -7518,6 +7594,28 @@ components:
...
+      type: object
+      properties:
+        rejectReason:
+          type: string
+          description: Reason provided by the validator for rejecting the retake request
+        status:
+          $ref: "#/components/schemas/RetakeExamStatus"
+      required:
+        - rejectReason
+        - status
```
kirisaki-vk: Huh ?

kirisaki-vk: what about `required: true` for each field ?

kirisaki-vk: Leave as is if it's the recomended way

doc/api.yml
```diff
@@ -7518,6 +7594,28 @@ components:
...
         status:
           $ref: "#/components/schemas/RetakeExamStatus"
+    CancelReason:
+      type: object
+      properties:
+        cancelReason:
+          type: string
+          description: Reason provided by the student for canceling the retake request
+        status:
+          $ref: "#/components/schemas/RetakeExamStatus"
```
kirisaki-vk: These are not needed as a request body for a cancel reason since it's a cancel request so you know the status will be TO_CANCEL, also doing this will have security risks

doc/api.yml
```diff
@@ -7518,6 +7594,28 @@ components:
...
+        - cancelReason
+        - status
+    RejectReason:
+      type: object
+      properties:
+        rejectReason:
+          type: string
+          description: Reason provided by the validator for rejecting the retake request
+        status:
+          $ref: "#/components/schemas/RetakeExamStatus"
```
kirisaki-vk: Same thing

kirisaki-vk (approved):

PR #1074, opened 2025-11-06
feat: add retake exam session student level

DyferHerioss:


src/main/java/school/hei/haapi/model/RetakeExamSession.java
```diff
@@ -41,4 +48,12 @@ public class RetakeExamSession implements Serializable {
...
   @JsonManagedReference
   private List<RetakeExam> retakeExams = new ArrayList<>();
+
+  @ElementCollection(targetClass = StudentLevel.class)
+  @CollectionTable(
+      name = "retake_exam_session_student_level",
+      joinColumns = @JoinColumn(name = "retake_exam_session_id"))
+  @Column(name = "\"student_level\"")
+  @JdbcTypeCode(NAMED_ENUM)
+  private List<StudentLevel> studentLevels = new ArrayList<>();
```
kirisaki-vk: Use `List.of()` if you need to provide a default value 
`ArrayList` is mutable

src/main/java/school/hei/haapi/model/RetakeExamSession.java
```diff
@@ -41,4 +48,12 @@ public class RetakeExamSession implements Serializable {
   @OneToMany(mappedBy = "session")
   @JsonManagedReference
   private List<RetakeExam> retakeExams = new ArrayList<>();
```
kirisaki-vk: Use `List.of()` if you need to provide a default value

kirisaki-vk (changes requested):

src/main/java/school/hei/haapi/model/RetakeExamSession.java
DyferHerioss: So I don't need a default value, I'm going to remove it

kirisaki-vk (approved):

PR #1084, opened 2025-11-11
feat: (doc) add new endpoint to approve or reject multiple request to cancel retake exam

DyferHerioss:


doc/api.yml
```diff
@@ -7631,22 +7658,24 @@ components:
...
           type: string
-          description: Reason provided by the student for canceling the retake request
-      required:
-        - cancelReason
-    RejectReason:
+        status:
+          $ref: "#/components/schemas/RetakeExamStatus"
+        approval_or_rejection_reason:
+          allOf:
+            - $ref: "#/components/schemas/Reason"
```
kirisaki-vk, line 7670: Isn't the reason only needed when the cancelation request is rejected ?

doc/api.yml
```diff
@@ -1598,6 +1598,39 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /retake_exams/approve_or_reject_cancel_retake_exams_request:
```
kirisaki-vk, line 1601: hmmmm.....
you need to change this name
This is more like update retake exam status

kirisaki-vk (approved):

PR #1145, opened 2026-01-13
fix: enable repeater grades for previous groups

tokyramarozaka:
This small fix is used to allow **adding grades for repeaters for their previous groups**. 

It used to be impossible, as they no longer belonged to that group, throwing an error saying that they are no longer a part of the groups assigned to that exam... yet, they _used_ to be.

### DIFF
--- a/src/test/java/school/hei/haapi/service/GradeServiceTest.java
+++ b/src/test/java/school/hei/haapi/service/GradeServiceTest.java
@@ -4,13 +4,22 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.when;
+import static school.hei.haapi.integration.test_data.CourseAssignmentTestData.createCourseAssignment;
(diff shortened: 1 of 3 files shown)


src/test/java/school/hei/haapi/service/GradeServiceTest.java
```diff
@@ -40,4 +49,27 @@ void exam_with_no_grades() {
 
     assertEquals("Exam with id " + examId + " do not have a score", exception.getMessage());
   }
+
+  @Test
+  void createParticipantGrade_forPreviousGroup_ok() {
```
YumeT023, line 71: I think a non regression test is required, a test that asserts that it effectively throws when trying to create a grade for a group a given student were never part of ... if not already done

YumeT023 (changes requested):

src/test/java/school/hei/haapi/service/GradeServiceTest.java
tokyramarozaka: added for both possible errors : group-less students, and students who were _never_ part of the assigned groups.

YumeT023 (approved):

PR #1146, opened 2026-01-13
feat: (doc) add new endpoint to download student results summary transcript

DyferHerioss:


doc/api.yml
```diff
@@ -622,6 +622,32 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /students/{student_id}/results_summary_transcript:
+    get:
+      tags:
+        - Grades
+      summary: Get the results summary of a given student
```
tokyramarozaka: be slightly explicit because result summary is a bit implicit : add something like `over all three / five years` or `1 page per year`

doc/api.yml
```diff
@@ -622,6 +622,32 @@ paths:
...
+      operationId: getResultsSummaryTranscript
+      parameters:
+        - $ref: "#/components/parameters/PathStudentId"
+      responses:
+        '200':
+          description: Results summary transcript of the specified student.
+          content:
+            application/vnd.openxmlformats-officedocument.spreadsheetml.sheet:
```
tokyramarozaka: PDF ?

tokyramarozaka (changes requested):

doc/api.yml
DyferHerioss: fixed

doc/api.yml
DyferHerioss: changed to more explicit summary

PR #1153, opened 2026-01-20
fix: add UNPAID_COUNT stat type to advancedFeeStats

tokyramarozaka:
AdvancedFeeStats is a class that counts : PAID_COUNT, PENDING_COUNT, LATE_COUNT, TOTAL_COUNT between two dates. 

**UNPAID_COUNT** (_fees that are not paid, and not late_)  however is missing, which makes it hard to verify how the sum of the statistics are computed. 

Therefore, I added both API specifications, tests, and additional service code to make it work. 

The previous code of AdvancedFeeStats in itself needs some refactoring as it is quite repetitive, but the advantage is that it simply required me to copy past the repetitive code for UNPAID_COUNT, and adjust.

### DIFF
--- a/doc/api.yml
+++ b/doc/api.yml
@@ -8698,6 +8698,8 @@ components:
           $ref: "#/components/schemas/PendingFeesStats"
         late_fees_count:
           $ref: "#/components/schemas/LateFeesStats"
+        unpaid_fees_count:
+          $ref: "#/components/schemas/UnpaidFeesStats"
         update_datetime:
           type: string
           format: date-time
@@ -8757,6 +8759,11 @@ components:
         - $ref: "#/components/schemas/BaseFeesStats"
         - $ref: "#/components/schemas/RetakeExamFeesCount"
     LateFeesStats:
+      type: object
+      allOf:
+        - $ref: "#/components/schemas/BaseFeesStats"
+        - $ref: "#/components/schemas/RetakeExamFeesCount"
+    UnpaidFeesStats:
       type: object
       allOf:
         - $ref: "#/components/schemas/BaseFeesStats"
--- a/src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
+++ b/src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
@@ -189,8 +189,7 @@ public class AdvancedFeeStatsService {
     feeStats.setUnknownFrequencyCount(pendingFeesStats.getUnknownFrequency());
   }
 
-  private void handleUnpaidFeesCount(AdvancedFeeStats feeStats, UnpaidFeesStats unpaidFeesStats) {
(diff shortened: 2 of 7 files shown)


doc/api.yml
```diff
@@ -8761,6 +8763,11 @@ components:
       allOf:
         - $ref: "#/components/schemas/BaseFeesStats"
         - $ref: "#/components/schemas/RetakeExamFeesCount"
+    UnpaidFeesStats:
+      type: object
+      allOf:
+        - $ref: "#/components/schemas/BaseFeesStats"
+        - $ref: "#/components/schemas/RetakeExamFeesCount"
```
YumeT023, line 8770: Just use one model for these stats, and keep the PaidFeesStats (extend the model and add the extra properties) so you don't have to write a lot of mapper code.

src/test/java/school/hei/haapi/service/AdvancedFeeStatsServiceTest.java
```diff
@@ -87,6 +90,34 @@ void getAdvancedFeeStats_shouldUpdateData_ok() {
     assertEquals(Boolean.TRUE, result.getExpired());
   }
 
+  @Test
+  void handleUnpaidFeesCount_copies_all_unpaidFeeStat_toAdvancedStat_ok() {
+    UnpaidFeesStats input =
```
YumeT023: use var when the type is ... obvious/too complicated

src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
```diff
@@ -186,6 +189,19 @@ private void handlePendingFeesCount(
     feeStats.setUnknownFrequencyCount(pendingFeesStats.getUnknownFrequency());
   }
 
+  public void handleUnpaidFeesCount(AdvancedFeeStats feeStats, UnpaidFeesStats unpaidFeesStats) {
```
YumeT023: Public visibility for this method is not appropriate given its intended use case. We typically use dedicated mapper classes for this type of operation. If this is exposed for testing purposes, the functionality can actually be tested through the service/controller endpoints that consume it. Exposing this method directly from the service doesn't follow our the architecture

YumeT023 (changes requested):

src/test/java/school/hei/haapi/service/AdvancedFeeStatsServiceTest.java
DyferHerioss: yes, I see

src/main/java/school/hei/haapi/service/AdvancedFeeStatsService.java
DyferHerioss: changed to package-private

PR #1170, opened 2026-02-17
feat(doc): add new endpoint to export all fee

DyferHerioss:


doc/api.yml
```diff
@@ -4012,14 +4012,53 @@ paths:
           $ref: '#/components/responses/429'
         '500':
           $ref: '#/components/responses/500'
+  /fees/export:
+    get:
+      tags:
+        - Paying
+      summary: Generate fees xlxs file
+      description: Generate an fees excel file, stores it in (S3 compatible storage), and returns a presigned URL taht can later be user to view the file.
```
tokyramarozaka: TYPO : * that

doc/api.yml
```diff
@@ -4012,14 +4012,53 @@ paths:
...
+        '429':
+          $ref: '#/components/responses/429'
+        '500':
+          $ref: '#/components/responses/500'
   /fees/advanced-stats/export:
     get:
       tags:
         - Paying
       summary: Generate advanced fees statistics file
       description: Generates an advanced fees statistics excel file, stores it in
```
tokyramarozaka, line 4060: what happens if the file already exists ? do we re-create the presigned url or do we update or what happens ?

doc/api.yml
```diff
@@ -4012,14 +4012,53 @@ paths:
...
+          $ref: '#/components/responses/403'
+        '429':
+          $ref: '#/components/responses/429'
+        '500':
+          $ref: '#/components/responses/500'
   /fees/advanced-stats/export:
     get:
       tags:
         - Paying
       summary: Generate advanced fees statistics file
```
tokyramarozaka: with an optional month range

doc/api.yml
```diff
@@ -4012,14 +4012,53 @@ paths:
...
+                type: string
+        '400':
+          $ref: '#/components/responses/400'
+        '403':
+          $ref: '#/components/responses/403'
+        '429':
+          $ref: '#/components/responses/429'
+        '500':
+          $ref: '#/components/responses/500'
   /fees/advanced-stats/export:
```
tokyramarozaka, line 4055: If we already have a sheets template to verify if the stats are coherent with the dashboard, why would we need to generate another advanced-stats sheets again ?

tokyramarozaka (changes requested):
Overall great, but I don't quite get why we need the advanced-stats endpoint.

doc/api.yml
DyferHerioss: There's no update actually, it always recreate another file even if it already existed

doc/api.yml
DyferHerioss: changed

doc/api.yml
DyferHerioss: added

doc/api.yml
DyferHerioss: This is our last endpoint that we supposed to change, but instead of modifying it, we'll create a new endpoint right now, and at the end if we won't really need this, we can delete it safely.

PR #1175, opened 2026-02-20
fix: scholar certificate outputs M1 being a fourth year instead of a separate MASTER cycle

tokyramarozaka:
Currently, every past student can get a scholar certificate saying : Fourth year or Fifth year in computer science, where as they only signed up for the Bachelor (Licence) and "fourth year" should be 
1) not possible to them, 
2) replaced by "1st year in Master cycle", fourth year is not a thing. 

The current code **does not care about registering for the Master cycle**, in fact, it **does not have a notion of Bachelor and Master cycle.** It just simply **computes the number of years that has passed to determine your level.** 

IMPORTANT : this causes the system to assume that _since you spent 3 years at HEI, automatically you're in L3_, yet... **grades still play a role in your true level.** 

The repeater issues will be taken care of in a separate PR.


tokyramarozaka:
Yes, PromotionLevelOutOfRange should be removed as the logic itself needs to be changed : level is also dependent on grades !

src/main/java/school/hei/haapi/model/promotion/PromotionLevelOutOfRange.java
```diff
@@ -1,6 +1,10 @@
 package school.hei.haapi.model.promotion;
 
 public class PromotionLevelOutOfRange extends RuntimeException {
```
YumeT023: Even though both extending `RuntimeException` and `ApiException` behave the same at runtime (see `InternalToRestException`), extending `ApiException` preserves the semantic hierarchy n keeps Api related exception under a common abstraction

src/main/java/school/hei/haapi/model/CycleLevel.java
```diff
@@ -0,0 +1,17 @@
...
+import school.hei.haapi.endpoint.rest.model.StudentLevel;
+
+@Getter
+public enum CycleLevel {
+  BACHELOR(List.of(StudentLevel.L1, StudentLevel.L2, StudentLevel.L3)),
+  MASTER(List.of(StudentLevel.M1, StudentLevel.M2));
+
+  private final List<StudentLevel> levels;
+
+  CycleLevel(List<StudentLevel> levels) {
```
YumeT023: can be replaced with @AllArgsConstructors

src/main/java/school/hei/haapi/model/CycleLevel.java
```diff
@@ -0,0 +1,17 @@
+package school.hei.haapi.model;
+
+import java.util.List;
+import lombok.Getter;
+import school.hei.haapi.endpoint.rest.model.StudentLevel;
+
+@Getter
+public enum CycleLevel {
+  BACHELOR(List.of(StudentLevel.L1, StudentLevel.L2, StudentLevel.L3)),
+  MASTER(List.of(StudentLevel.M1, StudentLevel.M2));
```
YumeT023: Enum variant can be imported static-ally to make it less verbose

YumeT023 (changes requested):

PR #1214, opened 2026-04-10
chore: change vola verification endpoint target

Tiavina-Andriamamivony:


src/main/java/school/hei/haapi/endpoint/rest/controller/FeeController.java
```diff
@@ -90,6 +92,9 @@ public List<Fee> getFeesByStudentId(
       @RequestParam PageFromOne page,
       @RequestParam("page_size") BoundedPageSize pageSize,
       @RequestParam(required = false) FeeStatusEnum status) {
+    // Verify pending mpbs before fetching fees so results reflect current vola state.
+    // Orchestrated here to avoid circular dependency (MpbsService -> FeeService).
+    mpbsVerificationService.verifyPendingMpbsForStudent(studentId);
```
tokyramarozaka, line 97: for the current fee or for all fees ? because I can see that there is a {fee_id} as path variable, why do we need it ?

tokyramarozaka (changes requested):
not much to say other than the one change on the endpoint.

src/main/java/school/hei/haapi/endpoint/rest/controller/FeeController.java
Tiavina-Andriamamivony: To be honest, it surprised me too (it was already like that). I just added a check for the student's pending Mpbs when that URL is called; I didn't want to modify the endpoint more than that.

tokyramarozaka (approved):

PR #1233, opened 2026-05-20
feat/get student overview

DyferHerioss:
- **feat: get student result overviews**
- **feat: add event to cruptade result overview**

### DIFF
--- a/src/main/java/school/hei/haapi/model/StudentResultOverview.java
+++ b/src/main/java/school/hei/haapi/model/StudentResultOverview.java
@@ -17,10 +17,7 @@ import java.math.BigDecimal;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
-import lombok.Getter;
 import lombok.NoArgsConstructor;
-import lombok.Setter;
-import lombok.ToString;
 import org.hibernate.annotations.JdbcTypeCode;
 import org.hibernate.type.SqlTypes;
 
@@ -30,17 +27,14 @@ import org.hibernate.type.SqlTypes;
 @AllArgsConstructor
 @Data
 @Builder
-@Getter
-@Setter
-@ToString(exclude = {"promotion", "student"})
 public class StudentResultOverview {
   @Id
   @GeneratedValue(strategy = IDENTITY)
   private String id;
 
   @ManyToOne(fetch = FetchType.LAZY)
-  @JoinColumn(name = "promotion_id", nullable = false)
-  private Promotion promotion;
+  @JoinColumn(name = "graduation_promotion_id", nullable = false)
+  private Promotion graduationPromotion;
 
   @OneToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "student_id", nullable = false)
--- a/src/main/java/school/hei/haapi/model/StudentResultOverview.java
+++ b/src/main/java/school/hei/haapi/model/StudentResultOverview.java
@@ -10,6 +10,7 @@ import jakarta.persistence.FetchType;
 import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.Id;
 import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
 import jakarta.persistence.OneToOne;
 import jakarta.persistence.Table;
 import java.math.BigDecimal;
@@ -19,6 +20,9 @@ import lombok.Data;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
+import lombok.ToString;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
 
 @Entity
 @Table(name = "\"student_result_overview\"")
@@ -28,12 +32,13 @@ import lombok.Setter;
 @Builder
 @Getter
 @Setter
+@ToString(exclude = {"promotion", "student"})
 public class StudentResultOverview {
   @Id
   @GeneratedValue(strategy = IDENTITY)
   private String id;
 
-  @OneToOne(fetch = FetchType.LAZY)
+  @ManyToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "promotion_id", nullable = false)
   private Promotion promotion;
 
@@ -48,7 +53,8 @@ public class StudentResultOverview {
   private BigDecimal obtainedCredits;
 
   @Enumerated(EnumType.STRING)
-  @Column(name = "status", nullable = false)
+  @Column(columnDefinition = "result_overview_status")
+  @JdbcTypeCode(SqlTypes.NAMED_ENUM)
   private ResultOverviewStatus status;
 
   @Column(name = "total_credits", nullable = false)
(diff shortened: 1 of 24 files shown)


src/main/java/school/hei/haapi/model/StudentResultOverview.java
```diff
@@ -0,0 +1,62 @@
...
+import lombok.Setter;
+import lombok.ToString;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
+
+@Entity
+@Table(name = "\"student_result_overview\"")
+@NoArgsConstructor
+@AllArgsConstructor
+@Data
```
hei-admin-devs, line 28: rm because you already have getter setter and toString, rather do an EqualsAndHashCode, and caution in case you may have an infinite loop on the equals.

src/main/resources/application.properties
```diff
@@ -4,7 +4,7 @@ sentry.logging.enabled=true
 sentry.logging.minimum-event-level=warn
 
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
-
+spring.jpa.properties.hibernate.format_sql=true
```
hei-admin-devs: rm

src/test/java/school/hei/haapi/integration/test_data/StudentResultOverviewTestData.java
```diff
@@ -0,0 +1,164 @@
...
+  public static Promotion promotionCD() {
+    return Promotion.builder()
+        .name("Promotion 2021-2022")
+        .ref("Alumni 2021 2022")
+        .creationDatetime(Instant.parse("2021-01-01T00:00:00Z"))
+        .cycleLevel(BACHELOR)
+        .build();
+  }
+
+  public static User studentA() {
```
hei-admin-devs: maybe reuse the student into the student test data (axel, tolojanahary, freeddy)

src/test/java/school/hei/haapi/integration/StudentResultOverviewIT.java
```diff
@@ -0,0 +1,159 @@
...
+            studentResultOverviewA, studentResultOverviewB,
+            studentResultOverviewC, studentResultOverviewD));
+    groupFlowRepository.deleteAll(List.of(groupFlowA, groupFlowB, groupFlowC, groupFlowD));
+    promotionRepository.deleteAll(List.of(promotionAB, promotionCD));
+    groupRepository.deleteAll(List.of(groupAB, groupCD));
+    studentRepository.deleteAll(List.of(studentA, studentB, studentC, studentD));
+  }
+
+  @Test
+  public void get_student_result_overviews_OK() throws ApiException {
```
hei-admin-devs: test with students outside of the promotion itself.

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
```diff
@@ -0,0 +1,100 @@
...
+                  .obtainedCredits(yearlyResult.getObtainedCredits())
+                  .weightedAverage(yearlyResult.getWeightedAverage())
+                  .status(ResultOverviewStatus.valueOf(yearlyResult.getStatus().toString()))
+                  .totalCredits(yearlyResult.getTotalCredits())
+                  .build();
+            })
+        .toList();
+  }
+
+  public Promotion getStudentPromotion(
```
hei-admin-devs: Confusion: does this get the promotion in which the student is in, or the one he is going out with ?

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
```diff
@@ -0,0 +1,100 @@
...
+      YearlyResult yearlyResult, List<Promotion> promotions, User student) {
```
hei-admin-devs: why do we need the yearly result to get the promotion ? 

also why the list of promotions ? can't we just go like : student.getGroup().getPromotion() ?

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
```diff
@@ -0,0 +1,100 @@
...
+    var group = student.getGroupFlows().getLast();
+    var studentActualLevel = userService.getStudentLevel(student.getId());
+    var promotionIndex =
```
hei-admin-devs: isn't there a relation between promotion and group ?

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
```diff
@@ -0,0 +1,100 @@
...
+        IntStream.range(0, promotions.size())
+            .filter(i -> promotions.get(i).getGroups().contains(group))
+            .findFirst()
+            .orElse(-1);
+    if (studentActualLevel != L3 && studentActualLevel != M2) {
+      return promotions.get(promotionIndex);
+    } else if (yearlyResult.getStatus().equals(VALIDATED)) {
```
hei-admin-devs: I am going back to my initial question up there: is this the promotion of the student, or the one he goes out with ? please use another name other than "getPromotion" because it is confusing.

hei-admin-devs (changes requested):
You have to clarify what `getPromotion` really does, by changing its name.

src/main/java/school/hei/haapi/model/StudentResultOverview.java
hei-admin-devs: removed

src/test/java/school/hei/haapi/integration/StudentResultOverviewIT.java
hei-admin-devs: meaning that there is an overview result, and there's a student who belong to a past promotion.

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
DyferHerioss: it get the promotion that the student is going out with

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
DyferHerioss: We need yearlyResult and the list of promotions because:

- yearlyResult allows us to know whether the student has validated their Bachelor’s degree and can graduate with their current promotion. Otherwise, the student will be transferred to the next promotion.
- The list of promotions is needed to determine the promotion that comes after the student’s current one for the transfer process.

We can only use student.getGroup().getPromotion() if we already know in which promotion the student will be placed.

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
DyferHerioss: Yes, but students can belong to the same group and the same promotion during the academic year, while not necessarily validating their Bachelor’s degree at the same time.

A student may fail to validate their degree and therefore need to be transferred to a new graduation promotion, even though their group and current promotion remain the same during the courses.

src/main/java/school/hei/haapi/service/StudentResultOverviewService.java
DyferHerioss: I used a new name

src/test/java/school/hei/haapi/integration/test_data/StudentResultOverviewTestData.java
DyferHerioss: applied

PR #1246, opened 2026-06-12
fix: intercept request body and filter out OPTIONS and HEAD requests

Mbomain:


src/main/java/school/hei/haapi/endpoint/UserActivityInterceptor.java
```diff
@@ -43,7 +43,8 @@ public void afterCompletion(
         log.debug("Anonymous request, no principal");
       }
       String body = null;
-      if (request instanceof ContentCachingRequestWrapper wrapper) {
+      var attr = request.getAttribute("cachedRequestWrapper");
+      if (attr instanceof ContentCachingRequestWrapper wrapper) {
         byte[] buf = wrapper.getContentAsByteArray();
```
DyferHerioss: var

PR #1263, opened 2026-07-09
Feat: add fees only mode filter

RantoniainaAmbi:
Groupe 6:
Description
This PR introduces a fees-only mode filter to the API.

Configuration
The filter's behavior is dynamic and is enabled by defining the configuration directly in your .env file.

Changes
* feat: Added the fees-only mode filter.
* test: Added unit tests to validate the filter's behavior, including edge cases 
  (feesOnly = false, case sensitivity, prefix boundaries).
* docs: Documented the FEES_ONLY resource.
* fix: Corrected prefix matching to avoid false positives — a URI like "/feesnotreal" 
  was previously treated as an allowed subpath of "/fees".
* refactor: Removed manual constructor using System.getenv(); now relies solely on 
  Spring's @Value injection per review feedback.

### DIFF
--- a/src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
+++ b/src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
@@ -7,6 +7,7 @@ import jakarta.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.util.List;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import org.springframework.web.filter.OncePerRequestFilter;
 
@@ -16,11 +17,7 @@ public class FeesOnlyFilter extends OncePerRequestFilter {
 
   private final boolean feesOnly;
 
-  public FeesOnlyFilter() {
-    this("true".equalsIgnoreCase(System.getenv("FEES_ONLY")));
-  }
-
-  public FeesOnlyFilter(boolean feesOnly) {
+  public FeesOnlyFilter(@Value("${FEES_ONLY:false}") boolean feesOnly) {
     this.feesOnly = feesOnly;
   }
 
--- a/src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
+++ b/src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
@@ -37,8 +37,8 @@ public class FeesOnlyFilter extends OncePerRequestFilter {
       return;
     }
 
-    String uri = request.getRequestURI();
-    boolean allowed = ALLOWED_PREFIXES.stream().anyMatch(uri::startsWith);
+    var uri = request.getRequestURI();
+    var allowed = ALLOWED_PREFIXES.stream().anyMatch(uri::startsWith);
 
     if (!allowed) {
       log.info("FEES_ONLY mode: blocked request to {}", uri);
--- a/src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
+++ b/src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
@@ -1,0 +1,55 @@
+package school.hei.haapi.endpoint.rest.security;
+
+import jakarta.servlet.FilterChain;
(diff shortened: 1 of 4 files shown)


Andry-Tokiniaina:
projet 1: groupe 6: MEP

src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
```diff
@@ -0,0 +1,53 @@
...
+      HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+      throws ServletException, IOException {
+
+    if (!feesOnly) {
+      filterChain.doFilter(request, response);
+      return;
+    }
+
+    String uri = request.getRequestURI();
+    boolean allowed = ALLOWED_PREFIXES.stream().anyMatch(uri::startsWith);
```
DyferHerioss: use var

DyferHerioss (changes requested):
Create a test for fees_only = false

src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private final boolean feesOnly;
+
+  public FeesOnlyFilter() {
+    this("true".equalsIgnoreCase(System.getenv("FEES_ONLY")));
+  }
+
+  public FeesOnlyFilter(boolean feesOnly) {
+    this.feesOnly = feesOnly;
+  }
```
DyferHerioss, line 22: Remove this manual constructor and the String comparison. Spring can inject and convert the FEES_ONLY environment variable directly into a boolean using @Value("${FEES_ONLY:false}"), so there's no need to call System.getenv() and compare it with "true" manually.

DyferHerioss (changes requested):

src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
RantoniainaAmbi: Done — using var now.

src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlyFilter.java
RantoniainaAmbi: Added 5 test cases covering feesOnly = false (all-pass behavior, empty/root URIs, filterChain invocation).
Removed — now using only the @Value-injected constructor.

RantoniainaAmbi:
All requested changes addressed individually above. Also found and fixed a prefix-matching bug while writing edge-case tests: uri.startsWith(prefix) could match unintended paths like /feesnotreal. Fixed with exact-match-or-subpath check. Ready for re-review.

src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
```diff
@@ -1126,11 +1120,6 @@ private BearerAuthFilter bearerFilter(RequestMatcher requiresAuthenticationReque
         (httpServletRequest, httpServletResponse, authentication) -> {});
     bearerFilter.setAuthenticationFailureHandler(
         (req, res, e) ->
-            // note(spring-exception)
-            // issues like when a user is not found(i.e. UsernameNotFoundException)
-            // or other exceptions thrown inside authentication provider.
-            // In fact, this handles other authentication exceptions that are
```
DyferHerioss, line 1159: why remove the existing comments ? Don't touch the comments already existed.

src/test/java/school/hei/haapi/unit/FeesOnlyFilterTest.java
```diff
@@ -0,0 +1,231 @@
...
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.mock.web.MockHttpServletResponse;
+import school.hei.haapi.endpoint.rest.security.FeesOnlyFilter;
+
+@ExtendWith(MockitoExtension.class)
+class FeesOnlyFilterTest {
+
+  @Mock private FilterChain filterChain;
+
+  @Test
```
DyferHerioss, line 25: boy scout pattern, use var

DyferHerioss (changes requested):

PR #1265, opened 2026-07-09
Feat/fees only

mandresy753:
<h1 align="center"> Groupe 20 : </h1>

## Description

- This PR implements the `feesOnly` feature to restrict access to specific endpoints when the application runs in fees-only mode.

-The feature introduces access rules to allow fees-related operations, their required dependencies , and required public endpoints while blocking unrelated functionalities.

- The initial implementation was based on an annotation and interceptor mechanism. Following the review feedback, the access control logic was moved to a centralized `SecurityFilterChain` configuration to make the rules more explicit and easier to maintain.

## Changes

- Added fees-only access control behavior
- Defined allowed endpoints when `feesOnly` mode is enabled
- Centralized fees-only security rules in `SecurityFilterChain`
- Removed annotation-based interceptor implementation
- Updated affected controllers and security configuration

## Motivation

Provide a maintainable way to enable a restricted application mode where only fees-related features remain accessible.


src/main/java/school/hei/haapi/endpoint/FeesOnlyInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+    public FeesOnlyInterceptor(@Value("${FEES_ONLY:false}") boolean feesOnly) {
+        this.feesOnly = feesOnly;
+    }
+    @Override
+    public boolean preHandle(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            Object handler) throws Exception {
+
+        String uri = request.getRequestURI();
```
DyferHerioss: use var instead of String

src/main/java/school/hei/haapi/endpoint/FeesOnlyInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+
+        if (!feesOnly || isAlwaysAllowed(uri)) {return true;}
+        if (!(handler instanceof HandlerMethod handlerMethod)) {return true;}
+        boolean allowed =
```
DyferHerioss: the same here

src/test/java/school/hei/haapi/FeesOnlyTrueIT.java
```diff
@@ -0,0 +1,60 @@
...
+
+        assertDoesNotThrow(
+                () -> api.getFees(null, null, null, null, null, null, 1, 10, false, null));
+    }
+
+    @Test
+    void non_fees_only_endpoint_is_blocked_when_fees_only_enabled() {
+        var api = new UsersApi(anApiClient(MANAGER1_TOKEN));
+
+        ApiException exception =
```
DyferHerioss: use var

DyferHerioss (changes requested):

DyferHerioss (commented):
Create a filter chain in the security config instead of using an annotation, because a filter in a security config is more explicit than an annotation and easier to maintain.

src/main/java/school/hei/haapi/endpoint/FeesOnlyInterceptor.java
mandresy753: Done ✅

src/main/java/school/hei/haapi/endpoint/FeesOnlyInterceptor.java
mandresy753: Done ✅

src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlySecurityConfig.java
```diff
@@ -0,0 +1,40 @@
...
+    this.feesOnlyUriMatcher = feesOnlyUriMatcher;
+  }
+
+  @Bean
+  @Order(1)
+  public SecurityFilterChain feesOnlyFilterChain(HttpSecurity http) throws Exception {
+    if (!feesOnly) {
+      http.securityMatcher(request -> false)
+          .authorizeHttpRequests(req -> req.anyRequest().denyAll());
+      return http.build();
```
DyferHerioss, line 29: why use order ?

src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
```diff
@@ -68,6 +69,7 @@ public AuthenticationManager authenticationManager() {
   }
 
   @Bean
+  @Order(2)
   public SecurityFilterChain configure(HttpSecurity httpSecurity) throws Exception {
```
DyferHerioss, line 73: the same here, why you use order ?

DyferHerioss (changes requested):

src/main/java/school/hei/haapi/endpoint/rest/security/SecurityConf.java
mandresy753: @Order(2) ensures this chain is evaluated only after the feesOnly chain, so it only handles requests that aren't blocked by the feesOnly restrictions.

src/main/java/school/hei/haapi/endpoint/rest/security/FeesOnlySecurityConfig.java
mandresy753: `@Order(1)` ensures this chain is evaluated first, so requests forbidden in `feesOnly` mode get blocked before the other chain (which accepts everything) can let them through.

